PR #56, opened 2021-04-17
CC analysis for FIPS, improved results, switches

mmstanone:
(and a lot of commits)


cc_analysis_on_fips.py
```diff
@@ -0,0 +1,10 @@
+from pathlib import Path
```
J08nY, line 1: This file likely needs a better name, why is it a separate file?

fips_oop_demo.py
```diff
@@ -24,7 +28,7 @@ def main(config_file, json_file, no_download_algs):
     # dset = FIPSDataset({}, Path('./fips_test_dataset'), 'small dataset', 'small dataset for keyword testing')
 
     # Load metadata for certificates from CSV and HTML sources
-    dset.get_certs_from_web(json_file=json_file, redo=True)
+    dset.get_certs_from_web(json_file=json_file, redo=False)
```
J08nY, line 31: This `redo` argument changes with each PR I feel like. Maybe make it a command-line argument?

sec_certs/analyze_certificates.py
```diff
@@ -368,9 +373,15 @@ def build_cert_references(filter_rules_group, all_items_found):
 
 
 def analyze_references_graph(filter_rules_group, all_items_found, filter_label):
```
J08nY, line 375: I thought that the FIPS code no longer relies on the older CC code. But now I see that it indeed does and it mixes with the CC code in a really weird way, where functions are used for both FIPS and CC and its really just an interwoven mess. This needs to be address, at some point we want to drop the old CC code right?

sec_certs/certificate.py
```diff
@@ -739,9 +751,14 @@ def remove_algorithms(self):
     @staticmethod
     def get_compare(vendor: str):
         vendor_split = vendor.replace(',', '') \
-            .replace('-', ' ').replace('+', ' ').replace('®', '').split()
+            .replace('-', ' ').replace('+', ' ').replace('®', '').replace('(R)', '').split()
         return vendor_split[0] if len(vendor_split) > 0 else vendor
 
+    def clean_empty_entries(self):
+        self.web_scan.algorithms = [x for x in self.web_scan.algorithms if x != {"Certificate": []}]
```
J08nY, line 758: Could the empty entries not be created at all? E.g. change the creation algorithm so that it does not pollute the algorithm list with these entries? This also created a bunch of dictionaries only to compare against them.

sec_certs/dataset.py
```diff
@@ -1010,32 +1028,49 @@ def color_check(current_key):
...
+                cert_list = self.certs[key].web_scan.connections
+
+            else:
+                cert_list = self.certs[key].processed.connections
+            return cert_list
+
         keys = 0
         edges = 0
 
         highlighted_vendor = 'Red Hat®, Inc.'
```
J08nY, line 1045: Should this be here?

J08nY (commented):
Overall OK, but I thought that the new FIPS code was not meant to use the older CC code, but this does so heavily and creates quite a mess with mixing the FIPS and CC stuff.

sec_certs/certificate.py
```diff
@@ -130,7 +130,7 @@ class WebScan(ComplexSerializableType):
         algorithms: Optional[List[Dict[str, str]]]
         tested_conf: Optional[List[str]]
         description: Optional[str]
-        mentioned_certs: Optional[List[str]]
+        mentioned_certs: Optional[List[Dict]]
```
adamjanovsky, line 133: would be nice to know the exact type of `Dict`, like `Dict[str, str]`

sec_certs/certificate.py
```diff
@@ -144,6 +144,7 @@ class WebScan(ComplexSerializableType):
         revoked_link: Optional[str]
         sw_versions: Optional[str]
         product_url: Optional[str]
+        connections: List
```
adamjanovsky, line 147: same as above

sec_certs/certificate.py
```diff
@@ -174,6 +175,7 @@ class PdfScan(ComplexSerializableType):
         cert_id: int
         keywords: Dict
         algorithms: List
+        connections: List
```
adamjanovsky, line 178: same as above

sec_certs/certificate.py
```diff
@@ -199,6 +201,7 @@ class Processed(ComplexSerializableType):
         keywords: Optional[Dict]
         algorithms: Dict
         connections: List
+        unmatched_algs: int
```
adamjanovsky, line 204: same as above

adamjanovsky (commented):
What exactly is a CC analysis for FIPS? I do not understand its purpose.

sec_certs/certificate.py
```diff
@@ -629,10 +634,10 @@ def parse_cert_file_common(text_to_parse: str, whole_text_with_newlines: str,
         return items_found_all, whole_text_with_newlines
 
     @staticmethod
-    def parse_cert_file(text_to_parse: str, algorithms: List[Dict]) \
-            -> Tuple[Optional[Dict], str]:
+    def parse_cert_file(text_to_parse: str) -> Tuple[Optional[Dict], str]:
```
adamjanovsky, line 637: can we specify the `Dict` type here as well?

sec_certs/certificate.py
```diff
@@ -664,21 +669,27 @@ def parse_cert_file(text_to_parse: str, algorithms: List[Dict]) \
         return items_found_all, text_to_parse
 
     @staticmethod
-    def analyze_tables(cert: 'FIPSCertificate') -> Tuple[bool, 'FIPSCertificate', List]:
+    def analyze_tables(tup: Tuple['FIPSCertificate', bool] ) -> Tuple[bool, 'FIPSCertificate', List]:
```
adamjanovsky, line 672: same as above

sec_certs/analyze_certificates.py
mmstanone: @petrs suggested this - he said there was no reason to have the same functionality implemented multiple times. We can probably discuss it together.

adamjanovsky (changes requested):
My greatest concern is related to the fact that I don't understand the purpose of this PR.  Apart from that, I was mostly pointing only towards type hints that could be made more specific.

sec_certs/dataset.py
mmstanone: Do you mean the `highlighted_vendor` part or the clumsy web/pdf/processed selection?

sec_certs/certificate.py
mmstanone: oh, I realized this was completely unnecessary, thanks, will rewrite

sec_certs/dataset.py
J08nY: I meant the hard-coded `highllighted_vendor` variable.

sec_certs/certificate.py
```diff
@@ -547,15 +552,15 @@ def match_web_algs_to_pdf(cert: 'FIPSCertificate') -> int:
             for web_alg in alg_list:
                 if ''.join(filter(str.isdigit, web_alg)) not in all_algorithms:
                     not_found.append(web_alg)
-        logger.error(
+        logger.info(
             f"For cert {cert.dgst}:\n\tNOT FOUND: {len(not_found)}\n"
             f"\tFOUND: {sum([len(a['Certificate']) for a in cert.web_scan.algorithms]) - len(not_found)}")
         logger.error(f"Not found: {not_found}")
```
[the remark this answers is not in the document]
mmstanone: My idea was that for logging level `info` the user might want to get more detailed information from the log, hence the `logging.info()` part is more verbose, whereas `logging.error()` just logs how many algorithms weren't matched.

However, logging level `info` triggers `logger.error()` too (if that's what you meant), so I'll probably delete both logging entries anyway, as we are only concerned with the result (to get statistics and graphs) and other analyses aren't as verbose as this one.

sec_certs/dataset.py
mmstanone: I moved it to an argument with default value of `'Red Hat®, Inc.'`. I used this value because RH certificates populate a lot of space in the resulting graph and it's nice to see the results without an "extra touch".

adamjanovsky:
@mmstanone changed the target-branch to `old-api`.

mmstanone:
okay, @J08nY @adamjanovsky everything should be resolved now.

PR #69, opened 2021-04-26
Adding the instruction on how to "export" data out of the container

KeleranV:
Edit the Readme to add the instructions about how to use the image in order to get the generated files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ The analysis can be extended in several ways:
  3. Help to fix problems in data extraction - some PDF files are corrupted, there are many typos even in certificate IDs...
 
 ## How to run the application with a Docker container
- 1. pull the image from the DockerHub repository : `sudo docker pull seccerts/sec-certs`
- 2. run `sudo docker run --volume ./processed_data:/opt/sec-certs/examples/debug_dataset -it seccerts/sec-certs`
+
+ 1. pull the image from the DockerHub repository : `docker pull seccerts/sec-certs`
+ 2. run `docker run --volume ./processed_data:/opt/sec-certs/examples/debug_dataset -it seccerts/sec-certs`
  3. All processed data will be in the `~/processed_data` directory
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,3 +70,8 @@ The analysis can be extended in several ways:
  1. Additional keywords can be extracted from PDF files (modify `cert_rules.py`)
  2. Data from `certificate_data_complete.json` can be analyzed in a novel way - this is why this project was concieved at the first place.
  3. Help to fix problems in data extraction - some PDF files are corrupted, there are many typos even in certificate IDs...
+
+## How to run the application with a Docker container
+ 1. pull the image from the DockerHub repository : `docker pull seccerts/sec-certs`
+ 2. run `sudo docker run sec-certs:latest --volume ~/processed_data:/opt/sec-certs/examples/debug_dataset`
```
adamjanovsky, line 76: is it necessary to run the docker with sudo? I would strongly suggest avoiding this at all costs.

adamjanovsky: Also, `:latest` is fetched automatically when calling `docker pull seccerts/sec-certs` so it's not needed in `docker run`

adamjanovsky: Also, if the repository `sec-certs` is not found locally, it will not be found on the DockerHub as name of user is missing. `docker run seccerts/sec-certs` is needed.

adamjanovsky (changes requested):

README.md
adamjanovsky: Lastly, I was getting errors on complaining about some missing items on path. The resolution is to put the container name only after the `--volume` parameter. The final solution could then be:

```
docker run --volume ./processed_data:/opt/sec-certs/examples/debug_dataset -it seccerts/sec-certs
```

which will also run the process interactively, most probably showing the log to the user. 

🎉

KeleranV: For what I've tried, sudo is needed in order to call docker, i should also add sudo for the pull.
Thanks for the debugging of the `--volume` position !

adamjanovsky (approved):

PR #75, opened 2021-05-11
Changes in CAVP webpages parsing, fixed bugs, tests, and everything from the last PR

mmstanone:
(a lot of commits, again)
As the CAVP search page changed, I had to change the parsing process - that's the main reason for this PR, as the FIPS functionality broke down.

Do we wanna keep the `analyze_certificates.py` changes, or do we just ignore the file and move it into `old-api`? The main FIPS functionality is not using it at all.

The tests are ~a bit sturdy~ kinda silly, I just wanted to test whether I didn't break anything big enough to notice, so I used dependency trees that I'm sure are parsed correctly and check if some changes didn't break everything.

Apart from that, it's only bugfixes and small changes.


sec_certs/certificate.py
```diff
@@ -177,6 +178,7 @@ class PdfScan(ComplexSerializableType):
         cert_id: int
         keywords: Dict
         algorithms: List
+        connections: List
```
adamjanovsky, line 181: `List[str]` I assume?

sec_certs/dataset.py
```diff
@@ -1210,18 +1258,33 @@ def split_alg(alg_string):
...
-                for i in range(0, len(elements_sliced), 2):
-                    alg_type, alg_id = split_alg(elements_sliced[i].text.strip())
-                    validation_date = elements_sliced[i + 1].text.strip()
-                    fips_alg = FIPSCertificate.Algorithm(alg_id, vendor, implementation, alg_type, validation_date)
-                    if alg_id not in self.certs:
-                        self.certs[alg_id] = []
-                    self.certs[alg_id].append(fips_alg)
+            tbody_contents = table.find('tbody').find_all('tr')
+            current_vendor = current_product = current_validation = current_date = ""
+            for tr in tbody_contents:
```
adamjanovsky, line 1263: The cyclomatic complexity of this function is really high, I suggest refactoring it. Also, consider using conditional expressions like
```python
current_validation = elem.text.strip() if 'validation-number' in attachment['id'] else ''
```

sec_certs/dataset.py
```diff
@@ -884,44 +892,50 @@ def convert_all_pdfs(self):
         ]
         cert_processing.process_parallel(FIPSCertificate.convert_pdf_file, tuples, constants.N_THREADS)
 
-    def get_certs_from_web(self, redo: bool = False, json_file: Optional[Path] = None):
+    def prepare_dataset(self, test: Optional[Path] = None):
```
adamjanovsky, line 895: All public methods in `CCDataset` have an optional argument `update_json: bool = True` that, just before the function ends, updates the dataset json, so that you don't have to invoke it manually as you do in `fips_oop_demo.py`. I suggest you follow the same approach for all public methods in your class.

sec_certs/dataset.py
```diff
@@ -1037,44 +1054,55 @@ def compare_certs(current_certificate: 'FIPSCertificate', other_id: str):
...
+            for rule in current_cert.processed.keywords['rules_cert_id']:
+                matches = current_cert.processed.keywords['rules_cert_id'][rule]
+                current_cert.processed.keywords['rules_cert_id'][rule] = [cert_id for cert_id in matches if
+                                                                          validate_id(current_cert,
+                                                                                      cert_id.replace('Cert.', '')
+                                                                                      .replace('cert.', '')
+                                                                                      .lstrip("#CA0 "))
+                                                                          and cert_id != current_cert.cert_id]
 
         for current_cert in self.certs.values():
```
adamjanovsky, line 1067: Quite high cyclomatic complexity. For a starter, I suggest you implement a `@staticmethod` that operates on certificate and refactor the method, it is far too complex.

sec_certs/dataset.py
```diff
@@ -1108,32 +1134,48 @@ def color_check(current_key):
                            (self.certs[current_key].web_scan.module_name
                             if self.certs[current_key].web_scan.module_name else ''))
 
+        def get_processed_list():
```
adamjanovsky, line 1137: Isn't it a code smell? Inner function to touch the global arguments implicitly? Anyway, these 9 lines could be rewritten as:

```python
def get_processed_list(connection_list: str, key:):
    attr = {'pdf': 'pdf_scan', 'web': 'web_scan', 'processed': 'processed'}[connection_list]
    return getattr(self.certs[key], attr).connections
```

adamjanovsky (changes requested):

adamjanovsky (approved):

PR #90, opened 2021-05-25
Workaround for MyBinder

### LINKED ISSUE #81: Dockerfile not working with MyBinder
We use the following instance on MyBinder.org: https://mybinder.org/v2/gh/crocs-muni/sec-certs/cc-feature-parity?filepath=notebooks%2Fcc_data_exploration.ipynb

Currently, MyBinder notebook will not start properly (although it will build the Dockerimage) because it is not compatible with our `Dockerfile` for unknown reasons. At the same time, once `DockerFile` is present in the root of repository, one cannot configure MyBinder any other way -- the Dockerfile will override all settings.

We should therefore do one of the following:
1. Adjust `Dockerfile` to start MyBinder properly: https://mybinder.readthedocs.io/en/latest/tutorials/dockerfile.html
2. Move Dockerfile elsewhere to not interfere with MyBinder

KeleranV:
Fixes #81 by doing two things:
- Move dockerfile to `docker` directory (adjust GitHub Actions file accordingly)
- add `runtime.txt` [configuration](https://mybinder.readthedocs.io/en/latest/using/config_files.html#runtime-txt-specifying-runtimes) saying that we want Binder to use Python 3.8


adamjanovsky:
Shouldn't we also modify GitHub Actions for docker build? Or will it find the right docker file automagically?

PR #94, opened 2021-06-09
Leo/bsi b soup

adamjanovsky:


adamjanovsky:
## As of 9 July, the following is done:

- [x] Traversal of BSI webpage, only active certificates

## As of 9 July, the following remains to be done:

- [ ] Full traversal of archived certificates
- [ ] Extraction of all information about a single certificate
- [ ] Feed the BSI outputs into `CommonCriteriaCert` class
- [ ] Output the mismatch between BSI source and CommonCriteria.org source
- [ ] Select 3 certificates and verify (using hash function) that their reports are the same

sec_certs/bsi_parse.py
```diff
@@ -136,23 +101,23 @@ class Bsitmp(ComplexSerializableType):
...
 
-subpage = Bsitmp(test_url)
-
-
-# print(subpage.certification_date)
-
-
-def process(root_url: str):
-    browser = BsiBrowser(root_url)
+def process(base_url: str):
```
adamjanovsky, line 119: It feels that this whole method should be part of `BsiBrowser.parse()` method.

sec_certs/bsi_parse.py
```diff
@@ -136,23 +101,23 @@ class Bsitmp(ComplexSerializableType):
...
+    browser = BsiBrowser(base_url)
     browser.parse()
     tmp_list = []
     for handler in browser.handler_list:
```
adamjanovsky, line 123: This whole for loop can be written as a list comprehension and you can call `tmp_list.extend([the_comprehension])`

sec_certs/bsi_parse.py
```diff
@@ -16,115 +16,80 @@
...
+
+        # ------------------------------------------------------------------------------------------------
+
+
+
+# To end the processing, a last class will be used with final links
+# to retrieve simple data written on the page
 
 
 class Bsitmp(ComplexSerializableType):
```
adamjanovsky, line 95: Name of class is not really explanatory. Should be named something like `BSICertificate`

sec_certs/bsi_parse.py
```diff
@@ -161,33 +126,15 @@ def process(root_url: str):
             temp = Bsitmp(url)
             # print(temp.html_content)
             tmp_list.append(temp)
-        for link in handler.archive_list:
-            arch_handler = BsiHandler(link)
+        for arch_handler in handler.handler_list:
+            for arch_url in arch_handler.link_list:
+                arch_temp = Bsitmp(arch_url)
+                tmp_list.append(arch_temp)
```
adamjanovsky, line 132: Same here, use comprehension and `.extend()` notation

sec_certs/bsi_parse.py
```diff
@@ -136,23 +101,23 @@ class Bsitmp(ComplexSerializableType):
...
-
-# print(subpage.certification_date)
-
-
-def process(root_url: str):
-    browser = BsiBrowser(root_url)
+def process(base_url: str):
+    browser = BsiBrowser(base_url)
     browser.parse()
     tmp_list = []
```
adamjanovsky, line 122: this could then be instance variable

sec_certs/bsi_parse.py
```diff
@@ -16,115 +16,80 @@
...
 
-"""
-To end the processing, a last class will be used with final links
-to retrieve simple data written on the page
-"""
+        self.handler_list = \
+            [BsiHandler("https://www.bsi.bund.de/" + a['href'])
+             for a in self.soup.find_all('a', href=True, recursive=True, title=re.compile('Archive'))]
+
+        if len(self.handler_list) != 0:
```
adamjanovsky, line 83: `if self.handler_list:`

adamjanovsky (commented):
Some intermediate review. Once done with your work, be sure to get rid of non-descriptive comments and make the code (at least somewhat) PEP-8 compliant.

sec_certs/bsi_parse.py
```diff
@@ -0,0 +1,140 @@
...
+class Bsitmp(ComplexSerializableType):
+    """
+    a class that will contain essentials data to compare to CC certs
+    """
+    html_content: list
+    certification_date: str
+    valid_until: str
+    soup: BeautifulSoup
+    id: str
+    pdf_links: list[str]
```
adamjanovsky, line 104: this won't work. You need to use `List[str]`

sec_certs/bsi_parse.py
```diff
@@ -0,0 +1,140 @@
...
+        self.link_list = []
+        self.handler_list = []
+        self.archive_list = []
+
+    def parse(self):
+
+        self.handler_list = \
+            [BsiHandler("https://www.bsi.bund.de/" + a['href'])
+             for a in self.soup.find_all('a', href=True, recursive=True)
+             if 'c-navigation-teaser' in str(a.get('class'))]
```
adamjanovsky, line 55: You can search directly by link class with `find_all(class='c-navigation-teaser')`

J08nY:
This is now done in #247.

PR #110, opened 2021-10-26
Compute dependencies

GeorgeFI:
Implemented functionality that computes dependencies for CCDataset class.


sec_certs/certificate/certificate.py
```diff
@@ -52,7 +52,7 @@ def to_json(self, output_path: Union[Path, str]):
     @classmethod
     def from_json(cls, input_path: Union[Path, str]):
         with Path(input_path).open('r') as handle:
-            return json.load(handle, cls=CustomJSONDecoder)
+            return json.load(handle, csl=CustomJSONDecoder)
```
adamjanovsky, line 55: @GeorgeFI Don't the tests fail because of this? The line should be `return json.load(handle, cls=CustomJSONDecoder)`. Otherwise I can't see why the tests are failing (everything else looks good to me). Provided that I have more insight, I'll try to run your branch this evening to see the problem.

sec_certs/certificate/common_criteria.py
```diff
@@ -194,6 +195,11 @@ class Heuristics(ComplexSerializableType):
         cert_lab: Optional[List[str]] = field(default=None)
         cert_id: Optional[str] = field(default=None)
 
+        directly_affected_by: Set = field(default=None)
+        indirectly_affected_by: Set = field(default=None)
+        directly_affecting: Set = field(default=None)
+        indirectly_affecting: Set = field(default=None)
```
adamjanovsky, line 201: The type should probably be `Optional[Set]`.

sec_certs/certificate/certificate.py
adamjanovsky: @GeorgeFI I took a look at the test and figured that the problem was already previously present in the code (and it was me who introduced it), sorry for that! The problem is that each subclass of `ComplexSerializableType` must have unique name. But when I introduced CPE matching to FIPS dataset, I created an attribute `Heuristics` which was already a `ComplexSerializableType` of CommonCriteriaDataset. 

I hotfixed the problem by renaming `Heuristics` types to `FIPSHeuristics`, `CCHeuristics` respectively. The longterm fix is probably include whole module path for serialization, e.g., `sec-certs.certificate.fips.heuristics` instead of `heuristics`. 

There still was one test not passing, caused by a typo in the template `toy_dataset.json`. I fixed that too 👍.

sec_certs/certificate/common_criteria.py
```diff
@@ -185,14 +185,18 @@ def cert_id(self) -> Optional[str]:
...
         cpe_matches: Optional[List[Tuple[float, CPE]]] = field(default=None)
         labeled: bool = field(default=False)
         verified_cpe_matches: Optional[Set[CPE]] = field(default=None)
         related_cves: Optional[List[CVE]] = field(default=None)
         cert_lab: Optional[List[str]] = field(default=None)
         cert_id: Optional[str] = field(default=None)
+        directly_affected_by: Optional[List] = field(default=None)
+        indirectly_affected_by: Optional[Set] = field(default=None)
+        directly_affecting: Optional[Set] = field(default=None)
+        indirectly_affecting: Optional[Set] = field(default=None)
```
adamjanovsky, line 199: It would be nice to have the type of the members of `List`, `Set` available. (Just guessing) `Optional[List[str]]` or `Optional[Set[str]]`?

tests/test_cc_heuristics.py
```diff
@@ -134,3 +134,14 @@ def test_protection_profiles_matching(self):
...
+    def test_single_record_dependency_heuristics(self):
+        # Single record in daset is not affecting nor affected by other records
+        heuristics = self.cc_dset['ebd276cca70fd723'].heuristics
+        self.assertEqual(heuristics.directly_affected_by, None)
+        self.assertEqual(heuristics.indirectly_affected_by, None)
+        self.assertEqual(heuristics.directly_affecting, None)
+        self.assertEqual(heuristics.indirectly_affecting, None)
+
+    def test_dependency_dataset(self):
+        assert True
```
adamjanovsky, line 147: I guess this is still unfinished?

sec_certs/model/dependency_finder.py
```diff
@@ -0,0 +1,117 @@
+from typing import List, Set, Dict, Tuple
+
+
+class DependencyFinder:
+
+    def __init__(self):
+        self.dependencies = {}
+
+    @staticmethod
+    def _update_direct_references(referenced_by: Dict, cert_id: str, this_cert_id: str) -> None:
```
adamjanovsky, line 10: General note to typehints (applicable on other places in the commits). Please, make the typehints as aspecific as possible. E.g. I assume that `referenced_by` has to be `Dict[str]` for the function to properly work (or any other type instead of `str`). If that is the case, please write `Dict[str]` instead of `Dict` to provide more guidance.

sec_certs/model/dependency_finder.py
```diff
@@ -0,0 +1,117 @@
+from typing import List, Set, Dict, Tuple
+
+
+class DependencyFinder:
```
adamjanovsky, line 4: Overall very well written. It does not exactly adhere to sklearn `BaseEstimator` (nor it inherits from it), but I guess it does not make sense to do that in this case.

I was not examining the algorithmical correctness of the computation, just the overall look and feel.

petrs: The "correctness" can be crosschecked against older version - happy to run it for you

adamjanovsky (approved):

PR #133, opened 2021-12-13
Typehints

mmstanone:
I tried to add typehints for the whole project - `mypy -p sec_certs --ignore-missing-imports` now finds no errors. (the flag just skips libraries without library stubs)
I'd look at mypy in CI later, I want to merge this so I won't have to do it all over again :smile: 

I found following problems:
- `certs` attribute of `Dataset` had to be "retyped" to `Mapping`, because we need it to be covariant (child classes of `Dataset` use child classes of `Certificate`). However, this brings in some problems:
  - `Mapping` is immutable, therefore setters (and indexing, because of `__setitem__()`) can't be used. In `__setitem__()` method in `Dataset` I converted the Mapping to a dict and called `__setitem__()` on that, couldn't find any other way.
  - When a class inherits from a child class of `Certificate` (i.e. `CCMaintenanceUpdatesDataset`), `certs` can't be typed, because dict is invariant - so I just `# type: ignore`'d them. If you know a better way, let me know.
  
  `certs` in child classes should not be `Mapping`s, because of the mutability problems. `MutableMapping` is also invariant.
- I mainly tried to silence mypy errors with conditions and raising exceptions - the messages are sometimes less descriptive, however, none of the scenarios should happen if the tool is used correctly (otherwise it crashes anyway). If you think the messages should be more descriptive, let me know.
- There are some `#TODO`s in the code, some are mine with specific problems (and questions about them). I would like to ask you to answer them (or fix the ones that are not mine :smile: )

There are probably more problems, so if anything seems wrong, let me know.


sec_certs/dataset/dataset.py
```diff
@@ -82,12 +82,14 @@ def __getitem__(self, item: str):
         return self.certs.__getitem__(item.lower())
 
     def __setitem__(self, key: str, value: 'Certificate'):
-        self.certs.__setitem__(key.lower(), value)
+        dict(self.certs).update({key.lower(): value})
```
J08nY, line 85: There is no way this will work right? You are creating a new dictionary object here and updating it, but this will do nothing to the inner `self.certs` object which will remain unchanged.

sec_certs/dataset/fips.py
```diff
@@ -135,8 +136,9 @@ def download_all_htmls(self) -> List[str]:
         failed = [c for c in failed if c]
 
         self.new_files += len(html_urls)
-        logger.info(f"Download failed for {len(failed)} files. Retrying...")
-        cert_processing.process_parallel(FIPSCertificate.download_html_page, failed, config.n_threads, progress_bar_desc="Downloading HTML files again")
+        if len(failed) != 0:
+            logging.info(f"Download failed for {len(failed)} files. Retrying...")
```
J08nY, line 140: `logger` not `logging`

sec_certs/sample/common_criteria.py
```diff
@@ -443,6 +449,8 @@ def download_pdf_target(cert: 'CommonCriteriaCert') -> 'CommonCriteriaCert':
             error_msg = f'failed to download ST from {cert.report_link}, code: {exit_code}'
             logger.error(f'Cert dgst: {cert.dgst}' + error_msg)
             cert.state.st_download_ok = False
+            if not cert.state.errors:
+                cert.state.errors = []
             cert.state.errors.append(error_msg)
```
J08nY, line 454: These three lines could likely be simplified right? Also not sure the initial check works correctly. If the `errors` attribute is missing it will raise an exception right? Are we sure the attr is there and just not sure it is a list? What is it then?

J08nY (changes requested):

sec_certs/sample/common_criteria.py
mmstanone: The problem is that `errors` is optional, therefore it can be `None`. Here we have to make sure it is a list, so we can append the error message. I think the `errors` attribute will be always present.

sec_certs/dataset/dataset.py
```diff
@@ -82,7 +82,9 @@ def __getitem__(self, item: str):
         return self.certs.__getitem__(item.lower())
 
     def __setitem__(self, key: str, value: 'Certificate'):
-        dict(self.certs).update({key.lower(): value})
+        aux = dict(self.certs)
```
J08nY, line 85: I would rather keep the original code here (the one before the PR) and make mypy ignore this part of code all together, this is convoluted.

J08nY (approved):

adamjanovsky (approved):

PR #141, opened 2021-12-15
Use billiard instead of concurrent.futures

J08nY:
This should work with celery now (the previous fix did not :slightly_frowning_face: ).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 wheel==0.34.2
 beautifulsoup4==4.9.3
+billiard==3.6.4.0
 certifi==2020.11.8
 chardet==3.0.4
 click==7.1.2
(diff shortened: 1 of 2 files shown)


adamjanovsky:
Just a question, did you empirically check that both threading and multiprocessing work with billiard? I see that some called methods changed (`r.result()` to `r.get()`, `r.done()` to `r.ready()`).

J08nY:
The tests pass and they use the threading part. I also tested the process pool part separately. The changes here are basically a revert of the concurrent.futures PR because the billiard API is the same as in multiprocessing. Actually threading in billiard is broken which is why for threading we use the classical multiprocessing ThreadPool.

adamjanovsky:
> The tests pass and they use the threading part. I also tested the process pool part separately. The changes here are basically a revert of the concurrent.futures PR because the billiard API is the same as in multiprocessing. Actually threading in billiard is broken which is why for threading we use the classical multiprocessing ThreadPool.

Merging then, but we should keep in mind this change, in case profiling (or just empirical evidence) would reveal lack of desired speed up.

PR #156, opened 2022-01-05
Fix from_dict ordering issue.

J08nY:
This moves from passing the dict data to the class as a tuple and instead uses `**kwargs`. Some fixes to the stored test data were also necessary to fix mistypes and a badly named key. The tes `fictional_cert` was changed to detect if this issue regresses.

### DIFF
--- a/sec_certs/sample/certificate.py
+++ b/sec_certs/sample/certificate.py
@@ -51,7 +51,7 @@ class Certificate(ABC, ComplexSerializableType):
     @classmethod
     def from_dict(cls: Type[T], dct: dict) -> T:
         dct.pop("dgst")
-        return cls(*(tuple(dct.values())))
+        return cls(**dct)
 
     def to_json(self, output_path: Optional[Union[str, Path]] = None):
         if output_path is None:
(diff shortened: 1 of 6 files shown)


adamjanovsky:
Great, thanks a lot. I only changed `assert` to `assertEqual` in test.

adamjanovsky (approved):

J08nY:
I intentionally use the base asserts as that is  how pytest is designed to be used: https://docs.pytest.org/en/6.2.x/assert.html
Using pytest together with unittest.TestCase  and its asserts is also possible but somewhat unnecessary. It also makes some plugins (pytest-monitor) not work. But it is a minor issue so we can keep what we have.

adamjanovsky:
> I intentionally use the base asserts as that is how pytest is designed to be used: https://docs.pytest.org/en/6.2.x/assert.html Using pytest together with unittest.TestCase and its asserts is also possible but somewhat unnecessary. It also makes some plugins (pytest-monitor) not work. But it is a minor issue so we can keep what we have.

Sorry about that, I wasn't aware of that. In that case, we should probably unify to `assert` across tests.

Issue #158, opened 2022-01-10
Normalize and connect matched certificate ids (missing year, versions, maintainance reports)


petrs:
For example, following certificates are logically connected:
BSI-DSZ-CC-0857 
BSI-DSZ-CC-0857-V2
BSI-DSZ-CC-0857-V2-2015_M01
and different example:
BSI-DSZ-CC-1059-2018
BSI-DSZ-CC-1059-V2-2019
BSI-DSZ-CC-1059-V3-2019
BSI-DSZ-CC-1059-V4-2021

PR #166, opened 2022-02-05
Regex expand

GeorgeFI:


sec_certs/sample/common_criteria.py
```diff
@@ -605,17 +630,46 @@ def extract_report_pdf_frontpage(cert: "CommonCriteriaCert") -> "CommonCriteriaC
...
+        response_nscib, cert.pdf_data.report_frontpage["nscib"] = helpers.search_only_headers_nscib(
+            cert.state.report_txt_path
+        )
+        response_niap, cert.pdf_data.report_frontpage["niap"] = helpers.search_only_headers_niap(
+            cert.state.report_txt_path
+        )
+        response_canada, cert.pdf_data.report_frontpage["canada"] = helpers.search_only_headers_canada(
+            cert.state.report_txt_path
+        )
+
```
adamjanovsky, line 642: Same comment as above applies.

sec_certs/sample/common_criteria.py
```diff
@@ -651,8 +705,129 @@ def compute_heuristics_cert_lab(self):
...
+            fixed_cert_id = CommonCriteriaCert._fix_anssi_cert_id(cert_id)
+
+        if CommonCriteriaCert._is_bsi_cert(cert_id):
+            fixed_cert_id = CommonCriteriaCert._fix_bsi_cert_id(cert_id, all_cert_ids)
+
+        if CommonCriteriaCert._is_spain_cert_id(cert_id):
+            fixed_cert_id = CommonCriteriaCert._fix_spain_cert_id(cert_id)
+
+        if CommonCriteriaCert._is_ocsi_cert_id(cert_id):
+            fixed_cert_id = CommonCriteriaCert._fix_ocsi_cert_id(cert_id)
```
adamjanovsky, line 831: Nitpicking: I would consider refactoring `_is_bsi_cert` and similar method to a single method: `get_cert_laboratory()` that would return `anssi`, `bsi`, `canada`, ... `unknown`. You could then define

```python
fix_method = {
    'anssi': CommonCriteriaCert._fix_anssi_cert_id
    'bsi': 'CommonCriteriaCert._fix_bsi_cert_id # <-- careful, you must actually use functools.partial here
    # and so on, you know what I mean
}
fix_method[self.get_cert_laboratory()](cert_id)
```

This is better expandable if we add something in the future, and it also has a single interface.

adamjanovsky (changes requested):
Overall great work, thanks for your effort. The logic is great. I had two types of comments:
- Just codestyle, refactoring, ...
- We should search for current use of `heuristics.cert_id` and get rid of it alltogether if possible. If it is used somewhere where it should not be replaced with `heuristics.normalized_cert_id`, we should just replace it with `pdf_data.cert_id` and delete the `heuristics` attribute anyway :)

The hard work is done, just some trivial refactoring remains and we can merge :)

sec_certs/sample/common_criteria.py
```diff
@@ -190,6 +190,7 @@ class CCHeuristics(ComplexSerializableType):
         related_cves: Optional[Set[str]] = field(default=None)
         cert_lab: Optional[List[str]] = field(default=None)
         cert_id: Optional[str] = field(default=None)
```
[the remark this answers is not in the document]
J08nY: Eh, changing the format like this creates issues for the site and all downstream users. I understand it is very much alpha stage software right now, but perhaps we can just replace the `cert_id` value with the new normalized one. It was a result of a heuristic applied to the cert ID matches anyway, so now it will just be a better/normalized heuristic, so that the field name can stay. One can always look at the `rules_cert_id` matches in the target or report and get additional info there.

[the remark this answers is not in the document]
adamjanovsky: Yes of course, @GeorgeFI, can we discard `normalized_cert_id` a keep `cert_id` in `CCHeuristics` object? Doing otherwise would break the webpage.

sec_certs/dataset/common_criteria.py
```diff
@@ -179,11 +179,10 @@ def from_web_latest(cls):
             return cls.from_json(dset_path)
 
     @property
-    def get_all_cert_id_references(self):
+    def all_cert_ids(self):
```
adamjanovsky, line 182: Can you check if the following code is equivalent to yours?

```python
@property
def all_cert_ids(self):
    all_cert_ids = set([x.heuristics.cert_id for x in self])
    all_cert_ids.update(itertools.chain.from_iterable([x.pdf_data.keywords_rules_cert_id for x in self]))
    all_cert_ids.update(
        itertools.chain.from_iterable(
            x.pdf_data.st_keywords["rules_cert_id"] for x in self if x.pdf_data.st_keywords
        )
    )
    all_cert_ids.discard(None)
    return all_cert_ids
```

Decide which one you like more, just suggesting a shorter variant (but yours probably runs faster). We have `itertools.chain.from_iterable()` on a bambillion of places in the project anyway...

sec_certs/sample/common_criteria.py
```diff
@@ -189,7 +198,7 @@ class CCHeuristics(ComplexSerializableType):
         verified_cpe_matches: Optional[Set[str]] = field(default=None)
         related_cves: Optional[Set[str]] = field(default=None)
         cert_lab: Optional[List[str]] = field(default=None)
-        cert_id: Optional[str] = field(default=None)
+        cert_id: Optional[str] = field(default=None)  # TODO - delete this
```
adamjanovsky, line 201: Please, see: https://github.com/crocs-muni/sec-certs/pull/166#discussion_r801129966
Sorry for the confusion :(

adamjanovsky (commented):
🥇

adamjanovsky:
@GeorgeFI not sure what was your discussion with @petrs, but this PR closes #158 ?

petrs:
@GeorgeFI @adamjanovsky correct, this PR shall close the https://github.com/crocs-muni/sec-certs/issues/158

adamjanovsky (approved):

PR #168, opened 2022-02-11
Leftover work with mypy and some random fixes

### LINKED ISSUE #86: CC: Better exception handling on pdf processing
When running `dset.extract_pdf_metadata()` on `CCDataset` , a `PyPDF2` package is called on multiple occassions. Its functions are failing here and there. There are some possible improvements of the current processing:

- [ ]  If the pdf is encrypted, `PyPDF2` attempts to override encryption to learn number of pages. If it fails, our whole function returns, ignoring `getDocumentInfo()` that needs not to fail. 
- [ ]  Many unnecessary warnings are displayed, like: `PdfReadWarning: Superfluous whitespace found in object header b'50' b'0' [pdf.py:1665]`
- [ ]  Some other issues (at least three more distinct are present) should be explored and if possible, the problems fixed.

### LINKED ISSUE #130: CVEDataset handling optimizations
Currently, `CVEDataset` object is required for both CPE matching, and for CVE matching task. When both tasks are completed in a sequence, we deserialize the `CVEDataset` object twice (which takes some time). Adding a logic that would keep the `CVEDataset` opened for subsequent use could save some computation time. 

Low priority probably, this just adds few seconds to comp. time.

### LINKED ISSUE #131: CC Leaking local paths in serialized errors
E.g. 

```
{'state': {'errors': ['Failed to read metadata of /Users/adam/phd/projects/certificates/datasets/cc_full_dataset/certs/targets/pdf/e6959c2b66202cb8.pdf,
```

Either add option to delete the errors (or state) once the dataset is processed or try to convert paths to relative.

### LINKED ISSUE #137: Leftover work with MyPY
Problems:

- [x] Typing of inherited types (Certificate -> CommonCriteriaCert)
- [x] TODOs in general
- [x] Dataset class does not inherit from ComplexSerializableType
- [x] FIPS Html table parsing reverted (does it work?)
- [x] Resolve `# type: ignore` stuff
- [x] Introduce `TyperVar`for handeling overloaded functions in child classes.

### LINKED ISSUE #169: Resolve set (de)serialization
Currently, sets that undergo process of serialization are sorted and converted to list during that process. Yet, when deserializing from JSON, they are kept as lists (as there is no way to tell whether these were originally lists or sets). 

We should (de)serialize sets as `ComplexSerializableType` to fix this, but it will create some backward compatibility problems, especially w.r.t. webpage. 

@J08nY will you survive this change? :)

adamjanovsky:
@mmstanone this is the PR we briefly talked about. I've put some `#TODO @sbobon` comments into the code for you (you better search for `@sbobon` in case I deviated from the outlined syntax). 

The fixes should be rather easy, but require some context on FIPS. 

## PR contents

- This closes #137.
- [1f929bf](https://github.com/crocs-muni/sec-certs/pull/168/commits/1f929bf678845e70caa6f95cd5899c9b1e4db494) also fixes #169.
- [ad667b4](https://github.com/crocs-muni/sec-certs/pull/168/commits/ad667b44561d2079b6c6ac428d213e3da0f983fb) fixes #131.
- [bce2bad](https://github.com/crocs-muni/sec-certs/pull/168/commits/bce2badb3f6b3577b23538cfef3fff653c871d7d) fixes #130.
- [25c91f9](https://github.com/crocs-muni/sec-certs/pull/168/commits/25c91f9e50a5f109436fe163638dd4d57060ad4d) fixes #86

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,9 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 21.12b0
+    rev: 22.1.0
     hooks:
       - id: black
-        language_version: python3
+        language_version: python3.8
         args: ['--check']
   - repo: https://github.com/pycqa/isort
     rev: 5.10.1
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@ exclude = '''
   | \.mypy_cache
   | \.tox
   | venv
+  | certsvenv
   | \.venv
   | _build
   | buck-out
--- a/sec_certs/config/configuration.py
+++ b/sec_certs/config/configuration.py
@@ -1,13 +1,13 @@
 import json
 from pathlib import Path
-from typing import Union
+from typing import Any, Union
 
 import jsonschema
 import yaml
 
 
 class Configuration(object):
-    def load(self, filepath: Union[str, Path]):
+    def load(self, filepath: Union[str, Path]) -> None:
         with Path(filepath).open("r") as file:
             state = yaml.load(file, Loader=yaml.FullLoader)
 
@@ -24,7 +24,7 @@ class Configuration(object):
         for k, v in state.items():
             setattr(self, k, v)
 
-    def __getattribute__(self, key):
+    def __getattribute__(self, key: str) -> Any:
         res = object.__getattribute__(self, key)
         if isinstance(res, dict) and "value" in res:
             return res["value"]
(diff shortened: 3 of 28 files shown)


adamjanovsky:
@mmstanone thanks for your contribution. Could you please fix the pipeline so that it passes lintng and tests?

mmstanone:
@adamjanovsky it should be done.

I also fixed issues that made flake8 fail in main branch, because it looked like some `#noqa`s went missing.

PR #171, opened 2022-02-16
Fix cpe matching inconsistencies

### LINKED ISSUE #170: Inconsistencies in CPE matching
Resolve several inconsistencies in CPE matching:

- [ ] CPEs are to be held in sets instead of lists
- [ ] CVEs are to be held in sets instead of lists
- [ ] When having >100 matches, "random" 100 matches are presented, decide if OK and document
- [ ] When having `includingEnd` (etc.) versions, random CPE is selected into the CPEDataset with `cpe.version == "-"`. This selection should be standardized (or all version ranges noted down) to keep CPEDatasets without excessive diffs
- [ ] `start_version` and `end_version` are not really utilized IIRC. But they are somehow processed in `CVEDataset`. We should check if by not utilizing them we omit some CVEs or not. If yes, we should document this in the code as this is highly unintuitive and tricky.
- [ ] Currently, if we (randomly) add some CPE with version interval into the CPEDataset, we don't actually use this interval during CPE matching (double-check). *However*, we use intervaled version as a key in dictionary of vulnerable CVEs, effectively discarding all other versions, this is kinda weird behaviour.

Example of weird diff in matched CPEs: https://gist.github.com/J08nY/f45ccd7974bc538c9d79d5b84ef69fd4

adamjanovsky:
Fixes #170

### DIFF
--- a/sec_certs/sample/common_criteria.py
+++ b/sec_certs/sample/common_criteria.py
@@ -12,7 +12,7 @@ from bs4 import Tag
 from sec_certs import constants as constants
 from sec_certs import helpers
 from sec_certs.model.cpe_matching import CPEClassifier
-from sec_certs.sample.certificate import Certificate, logger
+from sec_certs.sample.certificate import Certificate, Heuristics, logger
 from sec_certs.sample.protection_profile import ProtectionProfile
 from sec_certs.serialization.json import ComplexSerializableType
 from sec_certs.serialization.pandas import PandasSerializableType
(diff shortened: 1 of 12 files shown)


sec_certs/sample/common_criteria.py
```diff
@@ -770,6 +762,8 @@ def _fix_ocsi_cert_id(cert_id: str) -> str:
         return new_cert_id
 
     def get_cert_laboratory(self) -> str:
+        if not self.heuristics.cert_id:
+            raise ValueError("Cert ID was None but cert laboratory was to be computed based on its value.")
```
adamjanovsky, line 766: @GeorgeFI I needed to add these lines due to MyPy (it was not properly checking `heuristics` before, it does now). Is this ok with you?

Also, is there any specific reason why `heuristics.directly_affected_by` is a `list`, while other dependency attributes are `set`s? If possible, we would like to stick to sets to avoid unnecessary difs.

adamjanovsky:
@J08nY this should, to my best knowledge, fix all possible sources of inconsisentices in our CPE/CVE matching algorithms. Still, it created another problem (see #173), but that's a different story. I'll fix that soonish. Though, the code related to matching is quite complex due to the quality of JSONs distributed by NIST.

PR #177, opened 2022-02-22
[WIP] Unify handling of references in CC & FIPS.

### LINKED ISSUE #176: Source relationship data from both security target and report
In CC processing, the "affecting/affected" data is being sourced from the certificate report only (and not the security target). By doing this we are losing some interesting data. We would ideally want both data from the security target and the report separate in the certificate object and joined together in the same place as currently.

J08nY:
This will unify handling of references in CC and FIPS (and fixes #176 as well).
The current status is:

**CC**
Uses `DependencyFinder` to get the referenced and referencing certs from the certificate report only, then stores these in the `{in,}directly_{affecting,affected}` attributes of the heuristics object.

**FIPS**
Uses the `_find_connections` method of the FIPS dataset class to get the referenced certs from the security policy as well as the "Caveat" field from the web page, then stores the referenced certificates only in the `connections` attribute.

The ideal solution unifies the way these are stored as well as extends the data in the CC case to include the security target as a source of references as well.

### DIFF
--- a/sec_certs/dataset/common_criteria.py
+++ b/sec_certs/dataset/common_criteria.py
@@ -699,9 +699,11 @@ class CCDataset(Dataset[CommonCriteriaCert], ComplexSerializableType):
                 if not kws:
                     return set()
                 return set(kws["rules_cert_id"].keys())
+
             return func
+
         finder = DependencyFinder()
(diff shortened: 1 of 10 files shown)


adamjanovsky (approved):
pff, well, if it works :). The `setattr`, `getattr` stuff is quite wild.

PR #180, opened 2022-03-19
Dependency vulnerability

### LINKED ISSUE #147: Functionality: detect possibly vulnerable certificates by shared terms in certificate(s)
Example use-case: estID, ROCA and ID 163484 eIDAS memo. The memo referenced vulnerable chips, but a different ones than used in estID cards.  

Idea:
- input: list of known vulnerable certificates
- generate --find-affecting graph
- find term(s) shared by all known vulnerable certificates in their --find-affecting graphs
- generate graph with potentially affected certificates for these terms using --find-affected
-

GeorgeFI:
This closes #147.


sec_certs/sample/common_criteria.py
```diff
@@ -708,13 +708,18 @@ def _get_dependency_cves(self, dset: "CCDataset", dependency_type: DependencyCVE
...
 
         vulnerabilities = set()
+        dataset_cert_id_occurrences = dset.get_dataset_cert_ids_occurrences()
 
         for cert_id in references:
+            cert_id_occurrence = dataset_cert_id_occurrences.get(cert_id, None)
+
+            # Not in dataset or more than 2 occurrences in dataset -> skip
+            if cert_id_occurrence is None or cert_id_occurrence >= 2:
+                continue
```
adamjanovsky, line 718: Suggestion: Can we count the total number of thrown-away certificates and `log.warning()` this number?

sec_certs/dataset/common_criteria.py
```diff
@@ -684,12 +684,17 @@ def _compute_normalized_cert_ids(self):
         for cert in self:
             cert.compute_heuristics_cert_id(self.all_cert_ids)
 
+    def _compute_dependency_vulnerabilities(self):
+        for cert in self:
+            cert.find_certificate_dependency_vulnerabilities(self)
```
adamjanovsky, line 689: I think that we have unified interface for vulnerabilities and dependencies between CC and FIPS. Would it be possible to generalize this method and implement it into abstract `Dataset` class, so that it works both on FIPS and CC? While irrelevant for your thesis, I think that it's little extra work with nice benefit. 

Just one caveat: I think that we're using different objects for holding `cert_id` in the respective frameworks. Also, FIPS cert_ids are 100% accurate so no need to check for duplicites, etc.

sec_certs/sample/common_criteria.py
```diff
@@ -18,6 +19,10 @@
 from sec_certs.serialization.json import ComplexSerializableType
 from sec_certs.serialization.pandas import PandasSerializableType
 
+if TYPE_CHECKING:
+    from sec_certs.dataset.common_criteria import CCDataset
```
adamjanovsky, line 23: Hm, I haven't seen this one, nice.

sec_certs/sample/common_criteria.py
```diff
@@ -27,6 +32,11 @@
 }
 
 
+class DependencyCVEType(Enum):
+    DIRECT = "direct"
+    INDIRECT = "indirect"
```
adamjanovsky, line 37: *Nitpicking*: Shouldn't this be just `DepdendencyType`? How is it related to CVE?

sec_certs/sample/common_criteria.py
```diff
@@ -241,6 +251,8 @@ class CCHeuristics(Heuristics, ComplexSerializableType):
         cert_id: Optional[str] = field(default=None)
         st_references: References = field(default_factory=References)
         report_references: References = field(default_factory=References)
+        direct_vulnerabilities: Optional[Set[str]] = field(default=None)
```
adamjanovsky, line 254: **Thought:** Technically, this could (and should) replace `related_cves` attribute, because it holds exactly this information. I think that you are only expanding to "indirect" vulnerabilities, right?

sec_certs/sample/common_criteria.py
```diff
@@ -684,6 +696,41 @@ def compute_heuristics_cert_id(self, all_cert_ids: Set[str]):
         self.heuristics.cert_id = self.pdf_data.cert_id
         self.normalize_cert_id(all_cert_ids)
 
+    def _get_dependency_cves(self, dset: "CCDataset", dependency_type: DependencyCVEType) -> Optional[Set[str]]:
+        dependency_type_dict = {
+            "direct": self.heuristics.report_references.directly_referenced_by,
```
adamjanovsky, line 701: *Chore*: `"direct"` and `"indirect"` should be probably replaced with the `Enum`, right?

sec_certs/sample/common_criteria.py
```diff
@@ -684,6 +696,41 @@ def compute_heuristics_cert_id(self, all_cert_ids: Set[str]):
...
+            "indirect": self.heuristics.report_references.indirectly_referenced_by,
```
adamjanovsky, line 702: **Chore**: `"direct"` and `"indirect"` should be probably replaced with the `Enum`, right?

sec_certs/sample/common_criteria.py
adamjanovsky: Oh, I see. It does not! Because we are using 3 different states of vulnerabilities, right?
1. Certificate suffers from vuln.
2. Certificate *directly* references some other cert that suffers from vuln.
3. Certificate *indirectly* references some other cert that suffers from vuln.

Can we think of some better name please, so that 2 does not resemble 1, please?

sec_certs/sample/common_criteria.py
```diff
@@ -684,6 +696,41 @@ def compute_heuristics_cert_id(self, all_cert_ids: Set[str]):
...
+        for cert_id in references:
+            cert_id_occurrence = dataset_cert_id_occurrences.get(cert_id, None)
+
+            # Not in dataset or more than 2 occurrences in dataset -> skip
+            if cert_id_occurrence is None or cert_id_occurrence >= 2:
+                continue
+
+            for cert_obj in dset:
+                if cert_obj.heuristics.cert_id == cert_id:
+                    if cert_obj.heuristics.related_cves:
```
adamjanovsky, line 722: **Chore**: Rewrite as one if with `and` condition.

sec_certs/sample/common_criteria.py
```diff
@@ -684,6 +696,41 @@ def compute_heuristics_cert_id(self, all_cert_ids: Set[str]):
...
+                        vulnerabilities.update(cert_obj.heuristics.related_cves)
+
+        if not vulnerabilities:
+            return None
+
+        return vulnerabilities
```
adamjanovsky, line 728: **Chore:** rewrite as conditional expression: `return vulnerabilities if vulnerabilities else None`

sec_certs/sample/common_criteria.py
```diff
@@ -684,6 +696,41 @@ def compute_heuristics_cert_id(self, all_cert_ids: Set[str]):
...
+
+    def find_certificate_dependency_vulnerabilities(self, dset: "CCDataset"):
+        self.heuristics.direct_vulnerabilities = self._get_dependency_cves(dset, DependencyCVEType.DIRECT)
+        self.heuristics.indirect_vulnerabilities = self._get_dependency_cves(dset, DependencyCVEType.INDIRECT)
```
adamjanovsky, line 732: **Thought: It's slightly weird that a `CommonCriteriaCert` needs `CCDataset` class to compute this.

Can we maybe delete this method and perform the respective calls directly from the `CCDataset` object? I'm improvizing, but `CCDataset` class could have function roughly like

```Python
for cert in self:
    cert._get_dependency_cves(self, DependencyCVEType.DIRECT)
```

the `_get_dependency_cves` would need to be exposed as public though.

adamjanovsky: Naah, that doesn't help, it just moves the problem one step below, as `_get_dependency_cves()` should ideally not need the `CCDataset` variable as well...

Well, the most clean option would be to implement something in `model` module, similarly to the finder of references. That class would stand on top of both `CCDataset` and the certificates...

I understand that this would, however, require substantial refactoring. Leaving it to your consideration.

adamjanovsky (commented):
Hey, nice effort, I feel that this is important step of the analysis, as it could show us how fragile the ecosystem is, w.r.t. indirect vulnerabilities.

I was mostly nitpicking in my comments. But we should revisit the design of the classes. Right now, certificate objects need Dataset objects to make the computation. It's questionable whether we can get rid of this weird requirement, e.g. by making designated class (in `model` module) that would compute these vulns. for us.

sec_certs/dataset/common_criteria.py
```diff
@@ -749,6 +760,18 @@ def generate_cert_name_keywords(self) -> Set[str]:
         keywords.add("1.02.013")
         return {x for x in keywords if len(x) > config.minimal_token_length}
 
+    def get_dataset_cert_ids_occurrences(self) -> Dict[str, int]:
```
adamjanovsky, line 763: This is no more needed, right?

sec_certs/model/dependency_vulnerability_finder.py
```diff
@@ -0,0 +1,109 @@
...
+            for dgst in self.certificates:
+                cert_obj = self.certificates[dgst]
+                cert_obj_cves = cert_obj.heuristics.related_cves
+
+                if cert_obj.heuristics.cert_id == cert_id and cert_obj_cves:
+                    vulnerabilities.update(cert_obj_cves)
+
+        return vulnerabilities if vulnerabilities else None
+
+    def fit(self) -> Vulnerabilities:
```
adamjanovsky, line 75: *Suggestion**: The `__init__()` should not take certificates as its parameter. Instead, they should be supplied via `fit` method. 

If you take a look at how sklearn classifiers work, they are instantiated only with parameters. The data is filled into them in `fit()` method, that should also owerwrite the previous state. 

By doing so, you can actually reuse the `DependencyVulnerabilityFinder` with different lists of certificates.

sec_certs/model/dependency_vulnerability_finder.py
```diff
@@ -0,0 +1,109 @@
...
+            self.vulnerabilities[dgst][DependencyType.INDIRECT.value] = self._get_cert_dependency_cves(
+                dgst, DependencyType.INDIRECT
+            )
+
+        if thrown_away_cert_counter > 0:
+            logging.warning("There were total of %s certificates skipped due to duplicity", thrown_away_cert_counter)
+
+        return self.vulnerabilities
+
+    def get_dependency_vulnerabilities(self, dgst: str) -> DependencyCVE:
```
adamjanovsky, line 102: **Suggestion**: Rename to `predict_sigle_cert()`

we should also add a method `predict()` that would, given a list of certificates (or dgsts, idk), predict all the vulnerabilities. Again, this is to adhere with how sklearn classifiers typically work.

adamjanovsky: It would also be great if you could rename/create these methods for the reference finder that you prevously implemented. I missed it that time...

adamjanovsky (changes requested):
Hey, I've left some short remarks on how a member of `model` module should work. 

I would also love t see some new tests regarding this functionality. Trivial idea would be to check if the certificates from the tests have some indirect CVEs. If yes, we can manually verify those and save as templates to check against. If not, we must probably identify some suitable certificate. We should then add that one into CPE/CVE tests and invent a new test for finding indirect CVEs.

PR #183, opened 2022-03-31
Refactor/pdftotext as library

adamjanovsky:
@J08nY this may be of your interest w.r.t. deployment. 

`pdftotext` is no more invoked with `subprocess.run`, but is used as a library. The new way uses the same underlying binary, but something apparently changed and it causes the resulting text files to differ from the previous versions. I noticed:
- Some unicode characters (like trademark symbol) are now parsed properly
- Sometimes, spaces are now duplicated (i.e. two spaces instead of one appear in txt files)

Neither change should negatively impact the results. Still, we should beware the first run.

### DIFF
--- a/sec_certs/helpers.py
+++ b/sec_certs/helpers.py
@@ -3,7 +3,6 @@ import html
 import logging
 import os
 import re
-import subprocess
 import time
 from datetime import date, datetime
 from enum import Enum
@@ -14,6 +13,7 @@ from typing import Any, Dict, Generator, Hashable, Iterator, List, Optional, Seq
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+import pdftotext
 import pikepdf
 import requests
 from PyPDF2 import PdfFileReader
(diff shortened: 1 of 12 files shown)


tests/data/test_cc_oop/report_309ac2fd7f2dcf17.txt
```diff
@@ -1,16 +1,16 @@
-�rendetyp: 6 Diarienummer: 18FMV7705-43:1
+Ärendetyp: 6 Diarienummer: 18FMV7705-43:1
```
J08nY, line 1: Why did the id in this filename change? Is it the `dgst` of the certificate? If so, wy the change? Shouldn't that be static?

sec_certs/helpers.py
```diff
@@ -198,15 +198,20 @@ def repair_pdf(file: Path) -> None:
...
-    response = subprocess.run(
-        ["pdftotext", *options, pdf_path, txt_path], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, timeout=60
-    ).returncode
-    if response == 0:
-        return constants.RETURNCODE_OK
-    else:
+def convert_pdf_file(pdf_path: Path, txt_path: Path) -> str:
+    try:
+        with pdf_path.open("rb") as pdf_handle:
+            pdf = pdftotext.PDF(pdf_handle, "", True)
```
J08nY, line 204: Hopefully this is OK to be executed from the celery worker (and won't bring back issues addressed in #135 and #141). I can test it when the branch matures (ideally before merge).

J08nY:
One more question. Does this still leave the processed `.txt` files in the directories after processing? I am exposing these files on the website.

**Deploy notes:**

- [ ] Check that required libraries are installed on the server.
- [ ] Remove processed text files (otherwise the new text files will not overwrite the old ones). Or maybe change the logic to detect a file change?

adamjanovsky:
> One more question. Does this still leave the processed `.txt` files in the directories after processing? I am exposing these files on the website.
> 
> **Deploy notes:**
> 
>     * [ ]  Check that required libraries are installed on the server.
> 
>     * [ ]  Remove processed text files (otherwise the new text files will not overwrite the old ones). Or maybe change the logic to detect a file change?

Yes, this behaviour did not change.

adamjanovsky:
Making CI/CD work with new Poppler required some experimenting with plenty of commits that are now worth ammending, squashing, and force pushing. Instead, I will simply squash on merge, as this PR is otherwise quite compact.

PR #186, opened 2022-04-04
Refactor/better requirements

adamjanovsky:
See [CONTRIBUTING.md](https://github.com/crocs-muni/sec-certs/blob/e243d3cb782f1cad166be7b27ea38465352d0b05/CONTRIBUTING.md#requirements) for description.


adamjanovsky:
@J08nY this has a potential of breaking some things, as I got rid of requirements I couldn't identify and updated bunch of pinned dependencies along the way. I'll test it both locally and in CI/CD before merging, but ideally you should too...

Issue #173, opened 2022-02-17
Fix enhancing CPE dataset wtih CPEs only found in CVE database


adamjanovsky:
At the moment, CPE matches are too benevolent. This is because:

- When CPEDataset is being enhanced with CPEs specified *only* in CVEDataset, the record of `start_version`, `end_version` is deleted, ([see code](https://github.com/crocs-muni/sec-certs/blob/cbbb57f7097b9a61154d59bdd965811e0650befe/sec_certs/dataset/cpe.py#L177-L181)). 
- But then, the corresponding CPE in its full form (with `start_version`, `end_version`) is added to the matching dictionary, ([see code](https://github.com/crocs-muni/sec-certs/blob/cbbb57f7097b9a61154d59bdd965811e0650befe/sec_certs/dataset/cve.py#L225)).
- Subsequently, when look up dictionary CPE->CVE IDs is constructed, only `cpe.uri` [is checked](https://github.com/crocs-muni/sec-certs/blob/cbbb57f7097b9a61154d59bdd965811e0650befe/sec_certs/dataset/cve.py#L81), which is the same as in the CPE without `start_version`, `end_version`. 
- Example:

Imagine that we have certificate `Vendor: Product v16.5.7`. No similar CPE is present in the CPEDataset, but CPE `vendor:product:*:...` with `end_version: 8.1.0` is identified in CVEDataset. Then CPE `vendor:product:*:...` is added to the CPEDataset *without* the end version. Subsequently, if there's CVE concerning the CPEs under 8.1.0, our certificate would match despite being `16.5.7`.

Solution:

- The problem is that we consider CPE uri as a key to the dictionary. We therefore cannot register multiple CPEs that have different `start_version`, `end_version` fields.
- But during CPE matching we must keep theese start, end versions
- Solution may be: find suitable field in CPE that could carry this information so that it projects into the uri. Then create single CPE for each version pair.
- During matching, one would then check if the certificate version is in the desired interval.
- Hint: We could hijack `other` field in CPE. But we should check that all our CPEs in the dataset are of correct length. We should also check how often this field is utilized. Alternatively, version field could be hijacked as well.

Future work

- Immediately after issue is fixed, we should again check if all diffs are expected. If yes, then precision of CPE matching should be evaluated in label studio.

adamjanovsky:
[...]

PR #188, opened 2022-04-11
Improve CPE matching (again) and add SAR processing

adamjanovsky:
Finnish research of CVEs and present the results.

The list below represents current state:

### SARs and correlations

- [x] Double-check / verify the code for collecting SARs
- [ ] ~~Remains to collect both SARs and EALs from Protection Profiles~~
- [ ] ~~Rigorous leveling that adheres to CC standards should be in place (check for non-existant levels)~~
- [x] Rewrite  `Transformer` will be a better fit, also for DependencyFinder.
- [x] Shall we also collect SARs from reports?
- [x] Docstrings to the Estimator, check that it adheres to guidelines: https://scikit-learn.org/stable/developers/develop.html
- [x] Fix tests
- [x] Evaluate if conditions for SAR computation are ok
- [x] Add table to derive SARs from security level

*Note*: Level checks on SARs were discarded due to older versions of common criteria. It is insufficient to rewrite the newest CC specification into the form of regular expressions that are enforced. We must count also for older versions of security assurance components. We should not mind false positive as long as we compute correlations and enforce large support. 

### Further improvement of CVE matching

- [x] Disable enhancing CPEs with CPEs only existing in CVE database (temporary fix for #173)
- [x] Implement string lemmatization with Spacy prior to matching
- [x] Evaluate improvement of the measures outlined above


adamjanovsky:
### Comment on recent CPE matching advances

I did four things:
- Standardize version names in product names
- lower required match threshold from 100 to 90
- Forbid enhancing CPE datasets with CPEs that appear *only* in CVE database
- lemmatize product names, see below

Implementing lemmatization brings certificate product names into canonical form which simplifies the comparison with CPE strings. It would be time consuming to standardize all CPEs but I assume that they are mostly in the standard form. This is surely not the case for certificate names. Consider 

```python
"Infineon Security Controller M7892 B11 with optional RSA2048/4096 v1.02.013, EC v1.02.013, SHA-2 v1.01, SCL v2.02.012, Base v1.02.013, and Toolbox v1.02.013 libraries and with specific IC dedicated software (firmware)"
```
Direct comparison with CPE `"Infineon RSA Library 1.02.013"` yields `72` similarity score (due to `libraries` and `library` mismatch). We treat this by lemmatization that returns

```python
'infineon security controller m7892 b11 with optional rsa2048/4096 1.02.013 , ec 1.02.013 , sha-2 1.01 , scl 2.02.012 , base 1.02.013 , and toolbox 1.02.013 library and with specific ic dedicated software ( firmware )'
```

that gets `~92` similarity score.

All improvements yield the following results in comparison to previous version:
- Precision drop: 1%
- New certificates with >0 CPE match: 31
- Total number of new CPE matches: 588
- Not being too benevolent (see #173)

adamjanovsky:
@J08nY, prior to running sec-certs, a [Spacy](https://spacy.io/) language model must be downloaded with `python -m spacy download en_core_web_sm`.

PR #194, opened 2022-04-23
pdf to txt conversion with Raw=False

### LINKED ISSUE #185: Investigate  best pdf to txt conversion options
Historically, all pdf files were converted to text by calling `pdftotext` entry point of `poppler-utils` package (on Linux). This was (#183) replaced with [a pdftotext library](https://pypi.org/project/pdftotext/) that leveraged poppler API without a need to call `subprocess.open("pdftotext")`. 

Still, there is some parametrization that needs to be experimented with and settled. As conversion happen quite early in the pipeline, its quality can have paramount consequences on the final results. 

While [pdftotext library](https://pypi.org/project/pdftotext/) allows three different settings of `Raw` and `Physical` parameters, there are some other options directly in the `pdftotext` entry point that could possibly improve the quality of the conversion. 

Also, if some visually alligned choice is to be made, spaces need to get stripped before RE processing.

Note that `poppler-utils` must be installed to run `pdftotext` CLI.

adamjanovsky:
Closes #185

### DIFF
--- a/sec_certs/helpers.py
+++ b/sec_certs/helpers.py
@@ -201,7 +201,7 @@ def repair_pdf(file: Path) -> None:
 def convert_pdf_file(pdf_path: Path, txt_path: Path) -> str:
     try:
         with pdf_path.open("rb") as pdf_handle:
-            pdf = pdftotext.PDF(pdf_handle, "", True)  # No password, Raw=True
+            pdf = pdftotext.PDF(pdf_handle, "", False)  # No password, Raw=False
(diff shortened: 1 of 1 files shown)


J08nY:
This likely requires a change in the tests as there is a failing test that might be due to the pdf option change (and not NIST being down or something like that).

adamjanovsky:
Will be left with `Raw=True` for now.  Someone  can investigate closely #185 and reopen.

PR #199, opened 2022-04-26
Build wheels for sec-certs

adamjanovsky:


adamjanovsky:
Closing this MR. The fully self-contained wheel with bundled libraries (poppler, ...) won't be prepared. The user can use Docker for reproducible environment. Otherwise, the user should follow instructions that specify `poppler`, `graphviz` dependencies.

PR #206, opened 2022-05-12
populate __init__ for packages

### LINKED ISSUE #190: Aggregate to-be-imported modules into proper __init__.py file
Currently, when one wants to import e.g. `DependencyFinder` estimator, they must write:

```python
from sec_certs.model.dependency_finder import DependencyFinder
```

It would be wise to aggregate the classes into a module at one level higher, so that one can import more estimators from `model` module. E.g.,

```python
from sec_certs.model import DependencyFinder, CPEClassifier, ...
```

This holds also for `Dataset` and `Sample` modules.

adamjanovsky:
closes #190.

Also fixes some circular dependencies. Certificate objects now don't touch `CPEDataset` and `CVEDataset`, also `References` were moved from `model` module to `sample.Certificate`


J08nY:
Btw. instead of listing `__all__` manually which can get tedious, there is the [public](https://public.readthedocs.io/en/stable/) package (installed as `atpublic`). With it you can do:

```python
from public import public

@public
class Something:
  ...
```

and `Something` will automagically be added to `__all__`.

adamjanovsky:
> Btw. instead of listing `__all__` manually which can get tedious, there is the [public](https://public.readthedocs.io/en/stable/) package (installed as `atpublic`). With it you can do:
> 
> ```python
> from public import public
> 
> @public
> class Something:
>   ...
> ```
> 
> and `Something` will automagically be added to `__all__`.

magic. One of the reasons I've listed `__all__` is because of flake8 complaining about unused imports in `__init__.py` :smile:

PR #207, opened 2022-05-13
More notebooks

J08nY:


J08nY:
@adamjanovsky Do we want this? It is really a sketch that could be copied over into one of the quickstart notebooks for example, or ditched.

adamjanovsky:
I don't really care. As you say, maybe copy it here: https://github.com/crocs-muni/sec-certs/blob/main/notebooks/examples/common_criteria.ipynb ?

PR #209, opened 2022-05-13
warn user on missing poppler, graphviz

### LINKED ISSUE #208: Add user warnings in case of missing dependencies
If possible, users should be notified that they miss `graphviz` or `poppler` when running relevant parts of code. Those are:
- pdf conversion
- analysis

Notes:
- The checks are to be performed by CLIs. 
- The checks relies on heuristically obtained names of packages that may be platform dependent. 
- Checks require pkg-config. If not installed, the check will just raise a warning.

adamjanovsky:
closes #208

### DIFF
--- a/cc_cli.py
+++ b/cc_cli.py
@@ -9,6 +9,7 @@ import click
 
 from sec_certs.config.configuration import config
 from sec_certs.dataset import CCDataset
+from sec_certs.helpers import warn_if_missing_poppler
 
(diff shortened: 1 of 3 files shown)


adamjanovsky:
@J08nY I'm wondering how the warnings will work on Linux. They won't interrupt anything, but maybe you can check if you weren't warn by accident of missing `poppler` (CC) or `graphviz` (FIPS).

Issue #223, opened 2022-05-30
Add more variety to cert rules


J08nY:
Some ideas below:

## More device IDs
Maybe STM stuff, maybe ARM chip IDs?

## More crypto algs
Go through competitions and more weird algos:
 - AES candidates: http://competitions.cr.yp.to/aes.html
 - SHA3 competition: https://www.nist.gov/programs-projects/cryptographic-hash-algorithm-competition
 - PHC: https://www.password-hashing.net/
 - LWC: https://csrc.nist.gov/Projects/Lightweight-Cryptography
 - PQC: https://csrc.nist.gov/projects/post-quantum-cryptography
 - eSTREAM: http://competitions.cr.yp.to/estream.html
 - CAESAR: http://competitions.cr.yp.to/caesar.html

## More vendors and evaluation facilities
We have very few, there is definitely more, from the non-smartcard world for example.

https://www.commoncriteriaportal.org/labs/

## TEE-related stuff
TrustZone, SGX, etc.

## Improve the cert-id matching
Go through certificates that do not have an ID assigned to them and make or change a rule to match,
more normalization of the cert ID might be necessary.

adamjanovsky:
Not sure here, @J08nY, do you consider this thing resolved with #224 or do you want to continue? If yes, close it please.

J08nY:
This is fixed with #224, apart from the improved cert ID matching, which however is a long term goal separate from this issue.

Issue #14, opened 2020-11-19
Include search for specific APDU commands in documentation 


petrs:
Commands to obtain get proprietary chip info etc.
List of regular expressions are here:
https://github.com/petrs/pyAPDUFuzzer/commit/3cccb8ff72bf24cb20eb50208e45e563dff90a29
https://github.com/petrs/pyAPDUFuzzer/commit/d31dbd52a1833916d50ae678ca2fe101d0ada77e
https://github.com/petrs/pyAPDUFuzzer/commit/54f2060b85ba8dd398e1cc385d443b2b455113e7

adamjanovsky:
Won't be implemented, see https://github.com/crocs-muni/sec-certs/pull/224#issuecomment-1141907860

PR #224, opened 2022-05-30
Add more rules.

J08nY:
This moves towards #223.


J08nY:
@adamjanovsky This should make the tests fail yet it doesn't apparently we do not test full extraction out of a real certificate and compare the full output anywhere.

adamjanovsky:
@J08nY there's also a stale issue #14. Could you pls discuss with @petrs and possibly address this as well? At first sight it should be trivial, just add few more regexes (that Petr already defined).

J08nY:
I am worried about performance though. We can't really just go and add regexes willy-nilly. Each regex costs something and there seems to be like 100+ of the APDU ones and I am not sure they are worth it (@petrs wdyt?). I am currently trying to optimize our regex matching a bit, so compiled regexes are kept and they are not recompiled for each and every document. Without that, these 100 APDU regexes might be like 30 minutes more processing for the full run.

J08nY:
@adamjanovsky This should now be ready.

Btw. I think we should refactor the way we do rule matching a bit. What we have currently is a list of named groups where each group is a list of regex rules. The issue with that approach is that if you want to work on it and do data analysis on a more granular level than just the groups you need to go into the result matches and pick out individual regex results:

```python
"rules_symmetric_crypto": {
  "AES": 2,
  "AES-128": 3,
  "AES-256": 4
}
```
This way we actually lose the information that all of these are results from one regex and that that regex matches AES variants.
Our current rules structure also loses this info already, as we do not really have a policy on grouping things into one regex (only into rule groups). For example:

```python
"rules_symmetric_crypto" = [
  "AES",
  "AES-?(128|192|256)",
  ...
]
```
vs something new like

```python
"rules_symmetric_crypto" = {
  "AES": [
    "AES",
    "AES-?(128|192|256)"
   ],
  ...
}
```

adamjanovsky:
> I am worried about performance though. We can't really just go and add regexes willy-nilly. Each regex costs something and there seems to be like 100+ of the APDU ones and I am not sure they are worth it (@petrs wdyt?). I am currently trying to optimize our regex matching a bit, so compiled regexes are kept and they are not recompiled for each and every document. Without that, these 100 APDU regexes might be like 30 minutes more processing for the full run.

ok, closing #14 with `wont-fix`

adamjanovsky:
> @adamjanovsky This should now be ready.
> 
> Btw. I think we should refactor the way we do rule matching a bit. What we have currently is a list of named groups where each group is a list of regex rules. The issue with that approach is that if you want to work on it and do data analysis on a more granular level than just the groups you need to go into the result matches and pick out individual regex results:
> 
> ```python
> "rules_symmetric_crypto": {
>   "AES": 2,
>   "AES-128": 3,
>   "AES-256": 4
> }
> ```
> 
> This way we actually lose the information that all of these are results from one regex and that that regex matches AES variants. Our current rules structure also loses this info already, as we do not really have a policy on grouping things into one regex (only into rule groups). For example:
> 
> ```python
> "rules_symmetric_crypto" = [
>   "AES",
>   "AES-?(128|192|256)",
>   ...
> ]
> ```
> 
> vs something new like
> 
> ```python
> "rules_symmetric_crypto" = {
>   "AES": [
>     "AES",
>     "AES-?(128|192|256)"
>    ],
>   ...
> }
> ```

Ok, creating new issue #227 out of this. Could you please take a look at that? I don't see an easy solution atm. Alternatively we can discuss next week.

adamjanovsky (approved):

PR #226, opened 2022-06-02
Fix PyPi build

adamjanovsky:


adamjanovsky:
Also resolves https://github.com/crocs-muni/sec-certs/pull/225

PR #240, opened 2022-06-28
Reorganize rule matching

### LINKED ISSUE #227: Refactor matched regexes serialization
Currently, we lose information on the regexes that we've matched, see https://github.com/crocs-muni/sec-certs/pull/224#issuecomment-1141974642 for assignment.

J08nY:
Fixes #227.

This reorganizes the rule matching of both CC and FIPS in the following way:
 - Rules are no longer in a Python file but in a YAML file.
 - Rules have more than one level of categorization, for example the `cc_cert_id` rule has subcategories for each country issuing certificates and that subcategory has at least one but potentially more rules. This tree-like structure is passed on to the results of rule matching and is intentionally not flattened in the process.
 - The rule matching code is unified between FIPS and CC, and simplified somewhat.
 - The huge `helpers` module is split in several modules in the `utils` subpackage.

Overall, removing the effects of the Python -> YAML conversion and test dataset update (added many new lines), this change should be a net negative on the code size.

I suggest to squash this during merge, as the commits are not particularly of high quality.


sec_certs/rules.yaml
```diff
@@ -0,0 +1,1178 @@
+---
```
adamjanovsky, line 1: Seems ok. Please note that we must add this file into `package_data` in `setup.py`

adamjanovsky (commented):
Looks great, thanks for the effort. Just one note: I don't see atm how the changes affect `to_pandas()` serializations. Could you please try to inspect some random DataFrame and just check that all is ok? If not, then don't worry, I'll get to it quickly and fix it. 

Also, please add `yaml` files into `pakcage_data` in `setup.py`

sec_certs/rules.yaml
J08nY: Done. However, the way the other files are present in `package_data`, do they actually get included due to that? They are in subpackages and not in the root and the strings in the`package_data` do not have the subpackages.

J08nY:
The `to_pandas` serialization is unchanged as the pandas tuple is based on heuristics and those are all working with the new rule setup.

sec_certs/rules.yaml
adamjanovsky: I created a wheel (after you merged) with `python3 -m build .` and inspected its contents. All files specified in `package_data` are present in the wheel. So we should be safe.

Issue #213, opened 2022-05-20
FIPS: Fix from_web_latest()


adamjanovsky:
Currently, `from_web_latest()` of `FIPSDataset` is computing `finalize_results()` after loading the dataset from json. But that method is by the way computing CPEs, CVEs and stuff, which is super time consuming. There's some log message explaining why `finalize_results()` is needed, but its validity should be doublechecked. Ideally, only the minimal subset of actions should be taken as a part of that function.

Issue #211, opened 2022-05-19
FIPS: Broken graph plotting


adamjanovsky:
`FIPSDataset` currently cannot plot graphs with `plot_graphs()` function. Minimal working example obtained by fully processing the datasets, i.e. by calling

```python
dset = FIPSDataset()
dset.get_certs_from_web()
dset.convert_all_pdfs()
dset.pdf_scan()
dset.extract_certs_from_tables()
dset.finalize_results()
```

calling `dset.plot_graphs(show=False)` then fails with

```
(certsvenv) ~/phd/projects/certificates/sec-certs  $ /Users/adam/phd/projects/certificates/sec-certs/certsvenv/bin/python3 /Users/adam/phd/projects/certificates/sec-certs/examples/fips_oop_demo.py
Traceback (most recent call last):
  File "/Users/adam/phd/projects/certificates/sec-certs/examples/fips_oop_demo.py", line 49, in <module>
    main()
  File "/Users/adam/phd/projects/certificates/sec-certs/examples/fips_oop_demo.py", line 43, in main
    dset.plot_graphs(show=False)
  File "/Users/adam/phd/projects/certificates/sec-certs/sec_certs/dataset/fips.py", line 619, in plot_graphs
    self.create_dot_graph("full_graph", show=show)
  File "/Users/adam/phd/projects/certificates/sec-certs/sec_certs/dataset/fips.py", line 556, in create_dot_graph
    processed = self._get_processed_list(connection_list, key)
  File "/Users/adam/phd/projects/certificates/sec-certs/sec_certs/dataset/fips.py", line 521, in _get_processed_list
    return getattr(self.certs[dgst], attr).connections
AttributeError: 'FIPSHeuristics' object has no attribute 'connections'
```

Issue #210, opened 2022-05-19
FIPS: Refactor processing of Algorithms


adamjanovsky:
Currently, there's no public function for processing of `AlgorithmDataset` in FIPS, but a boolean flag that invokes the private calls instead. As the only thing that happens with algorithm dataset is `get_certs_from_web()`, they should be processed in similar fashion to CC Maintenance updates. In exact, there should be no boolean flags `not_download_algs` but a single public function `process_algorithms()` that handles the logic and assigns to `dset.algorithms`. 

The public function should be invoked by separate action `algorithms` from `fips_cli`.

PR #246, opened 2022-07-09
Harmonize FIPS code

J08nY:
This will be an overall FIPS improvement drive to get it to be more similar to CC code.

- [x] Add TODOs.
- [x] Make JSON serialization include full path to class instead of only classname.
- [x] Rename internal FIPSCertificate classes to match CC.
- [x] Remove unused methods.
- [x] Add PDF metadata extraction.
- [x] Make algorithm extraction (and serialization) deterministic.
- [x] #210
- [x] #213
- [x] #211
- [x] Split cert_id match cleaning to separate data structures, do not touch the keywords matches directly.

Possible future work:

- [ ] Also parse the detailed algorithm pages, such as [this one](https://csrc.nist.gov/projects/Cryptographic-Algorithm-Validation-Program/details?source=C&number=423). This is currently not done and only the search results page is scraped for the `FIPSAlgorithm` dataset. This significantly lowers the amount of data we collect.

### DIFF
--- a/fips_cli.py
+++ b/fips_cli.py
@@ -20,7 +20,18 @@ logger = logging.getLogger(__name__)
     required=True,
     nargs=-1,
     type=click.Choice(
-        ["new-run", "all", "build", "convert", "update", "pdf-scan", "table-search", "analysis", "graphs"],
+        [
+            "new-run",
+            "all",
+            "build",
+            "convert",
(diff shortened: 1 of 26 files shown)


J08nY:
Merging, even though the tests fail since it is only a NIST download fail.

PR #261, opened 2022-09-19
Do not reextract stuff in CC.

J08nY:
This phase of the extraction process is deterministic so we simply lose time doing this.


J08nY:
Tests are messed up externally, merging.

PR #285, opened 2022-11-27
Modelling "and" type CPEs with CPEConfiguration object

GeorgeFI:


sec_certs/dataset/cve.py
```diff
@@ -147,9 +147,35 @@ def from_json(cls, input_path: Union[str, Path]):
             dset = json.load(handle, cls=CustomJSONDecoder)
         return dset
 
-    def get_cve_ids_for_cpe_uri(self, cpe_uri: str) -> Optional[Set[str]]:
+    def _get_cve_ids_for_cpe_uri(self, cpe_uri: str) -> Optional[Set[str]]:
         return self.cpe_to_cve_ids_lookup.get(cpe_uri, None)
```
adamjanovsky, line 151: What would be the consequences of returning empty set by default instead of None?

sec_certs/dataset/cve.py
```diff
@@ -147,9 +147,35 @@ def from_json(cls, input_path: Union[str, Path]):
...
 
+    def _get_cves_from_exactly_matched_cpes(self, cpe_matches: set[str]) -> set[str]:
+        cve_ids: set[str] = set()
+
+        for cpe_match in cpe_matches:
+            if cve_set := self._get_cve_ids_for_cpe_uri(cpe_match):
+                cve_ids.update(cve_set)
+
+        return cve_ids
```
adamjanovsky, line 160: Nitpicking: Can we have it without the for loop? 😆 If we address the note above, then we could write

```python
def _get_cves_from_exactly_matched_cpes(self, cpe_matches: set[str]) -> set[str]:
    return set(itertools.chain.from_iterable([self._get_cve_ids_for_cpe_uri(x) for x in cpe_matches]))
```

sec_certs/dataset/cve.py
```diff
@@ -147,9 +147,35 @@ def from_json(cls, input_path: Union[str, Path]):
...
+        cves: set[str] = set()
+        cves_with_vulnerable_configurations = [cve for cve in self if cve.vulnerable_cpe_configurations]
+
+        for cve in cves_with_vulnerable_configurations:
+            if any([cpe_configuration.match(cpe_matches) for cpe_configuration in cve.vulnerable_cpe_configurations]):
+                cves.add(cve.cve_id)
+
+        return cves
+
+    def get_cves_from_matched_cpes(self, cpe_matches: set[str]) -> set[str]:
```
adamjanovsky, line 172: While this is premature, please comment the methods a bit before merging. This one is crucial to grasp, as it explains that we're matching using two different ways.

sec_certs/sample/cpe.py
```diff
@@ -8,6 +8,15 @@
 from sec_certs.utils import helpers
 
 
+class CPEConfiguration:
+    def __init__(self, platform: "CPE", cpes: list["CPE"]) -> None:
+        self.platform: "CPE" = platform
+        self.cpes: list["CPE"] = cpes
```
adamjanovsky, line 14: Do we actually need to hold the whole `CPE` object here? 

In the function below, you run potentially costly `map()` operation every time we try to match something. If we stored only `uri` instead of the whole CPE, we could get rid of this operation. Would that be possible?

adamjanovsky: Also, I'm undecided, but I'm thinking whether we should wrap all `CPE`s into this `CPEConfiguration` object, because technically a single CPE is a configuration as well. It would create some overhead, but semantically we would get rid of the schizophreny when we sometimes match `CPE` and sometimes `CPEConfiguration`. Not sure, what do you think?

sec_certs/sample/cve.py
```diff
@@ -66,11 +67,18 @@ def from_nist_dict(cls, dct: Dict[str, Any]) -> "CVE.Impact":
...
+        vulnerable_cpes: List[CPE],
+        vulnerable_cpe_configurations: list[CPEConfiguration],
+        impact: Impact,
+        published_date: str,
+        cwe_ids: Optional[Set[str]],
     ):
         super().__init__()
         self.cve_id = cve_id
         self.vulnerable_cpes = vulnerable_cpes
+        self.vulnerable_cpe_configurations = vulnerable_cpe_configurations
```
adamjanovsky, line 81: This is exactly the schizophreny I'm referring to. We have two objects: `vulnerable_cpes` and `vulnerable_cpe_configuraitons`. We could wrap all `CPE`s into `CPEConfiguration` I suppose.

sec_certs/sample/cpe.py
GeorgeFI: @adamjanovsky I can easily refactor the code to just store CPE URI, however, I wanted to create a factory method for this class that would check if any of the CPEs in `self.cpes` list is vulnerable. If there is any vulnerable CPE, I would create the `CPEConfiguration` object. In case there would be no vulnerable CPE, I would return `None` or throw some exception.

sec_certs/dataset/cve.py
```diff
@@ -147,9 +147,35 @@ def from_json(cls, input_path: Union[str, Path]):
...
+
+        for cpe_match in cpe_matches:
+            if cve_set := self._get_cve_ids_for_cpe_uri(cpe_match):
+                cve_ids.update(cve_set)
+
+        return cve_ids
+
+    def _get_cves_from_cpe_configurations(self, cpe_matches: set[str]) -> set[str]:
+        cves: set[str] = set()
+        cves_with_vulnerable_configurations = [cve for cve in self if cve.vulnerable_cpe_configurations]
```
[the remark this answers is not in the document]
GeorgeFI: Agree

sec_certs/sample/cpe.py
GeorgeFI: In the end, my approach with the factory method is not needed anymore. I solved that with `parse_only_vulnerable_cpes` parameter in `_parse_nist_dict` function.

adamjanovsky:
@GeorgeFI tests should work with this workflow file: https://github.com/crocs-muni/sec-certs/blob/be7f6d723368fcf2c7e5b4caf1e91de2a111531d/.github/workflows/tests.yml

adamjanovsky:
This is replaced by #300

PR #292, opened 2022-12-08
Add codecov

J08nY:


adamjanovsky:
@J08nY Can this be incorporated also into local development? I.e. can we create a pre-commit task for this? It's uncomfortable to discover failing pipelines only when they get to remote.

J08nY:
Well you can track code coverage when running tests. I do not think it makes sense to include anything based on that to a pre-commit. Otherwise every commit would run the test suite, I think that should not happen and be separate. If you want to look at the coverage diff you can check codecov manually, or just push and see the pipeline results.

PR #300, opened 2022-12-16
Feat cpe configurations

GeorgeFI:


src/sec_certs/dataset/cve.py
```diff
@@ -139,8 +147,35 @@ def from_web(
...
+        """
+        Method returns the set of CVEs which are matched to the set of CPEs.
+        First are matched the classic CPEs to CVEs with lookup dict and then are matched the
+        'AND' type CPEs containing platform.
+        """
+        cves = self._get_cves_from_exactly_matched_cpes(cpe_matches)
+        cves_matched_by_configurations = self._get_cves_from_cpe_configurations(cpe_matches)
+        cves.update(cves_matched_by_configurations)
+
+        return cves
```
adamjanovsky, line 178: Just `return {**self._get_cves_from_exactly_matched_cpes(cpe_matches), **self._get_cves_from_cpe_configurations(cpe_matches)`

src/sec_certs/sample/cve.py
```diff
@@ -48,11 +48,12 @@ def from_nist_dict(cls, dct: dict[str, Any]) -> CVE.Impact:
 
     cve_id: str
     vulnerable_cpes: list[CPE]
+    vulnerable_cpe_configurations: list[CPEConfiguration]
```
adamjanovsky, line 51: Just noticed this. Is there a reason that we have this as a list? Could be set as well, right?

src/sec_certs/sample/cve.py
```diff
@@ -110,74 +118,116 @@ def to_dict(self) -> dict[str, Any]:
...
-                itertools.chain.from_iterable(get_vulnerable_cpes_from_node(x) for x in dct["configurations"]["nodes"])
+            cpes_and_cpe_configurations = [
+                get_vulnerable_cpes_and_cpe_configurations(x, [], []) for x in dct["configurations"]["nodes"]
+            ]
+            vulnerable_cpes = list(itertools.chain.from_iterable(map(lambda x: x[0], cpes_and_cpe_configurations)))
+            vulnerable_cpe_configurations = list(
+                itertools.chain.from_iterable(map(lambda x: x[1], cpes_and_cpe_configurations))
             )
 
+            return vulnerable_cpes, vulnerable_cpe_configurations
```
adamjanovsky, line 222: Just `return [list(t) for t in zip(*cpes_and_cpe_configurations)]`

src/sec_certs/dataset/cve.py
```diff
@@ -55,9 +55,15 @@ def __len__(self) -> int:
     def __eq__(self, other: object):
         return isinstance(other, CVEDataset) and self.cves == other.cves
 
+    def _filter_cves_with_cpe_configurations(self) -> None:
+        """
+        Method filters the subset of CVEs, which contain at least one CPE configuration.
+        """
+        self.cves_with_vulnerable_configurations = [cve for cve in self if cve.vulnerable_cpe_configurations]
```
adamjanovsky, line 62: This should be set to empty list in `__init__`, so that we're explicit about its existance.

adamjanovsky (changes requested):
Hey,

thanks for your work. I'm happy that we're now on par with the main branch and matching the complex CPE configurations. Looking at the code, there are still some minor chunks to work on:

- We could have a test (similar to `test_find_related_cves()` in [test_cc_analysis.py](https://github.com/crocs-muni/sec-certs/blob/b5b0e9d1798338dca205b001e7b38f02c8929a89/tests/cc/test_cc_analysis.py#L118-L124) that would add some artifical CVE (feel free to make up your CVE, name of certificate and cert's CPEs) and check that those complex CVEs are actually properly matched.  Same test could be invented for FIPSDataset
- We're now fairly considerative when it comes to performance. Could you please measure how much memory does the old CPE/CVE matching takes and how much does the new one takes? Same for runtimes?
- I see that some optimizations could still be made. For instance [when building look-up dicts](https://github.com/crocs-muni/sec-certs/blob/b5b0e9d1798338dca205b001e7b38f02c8929a89/src/sec_certs/dataset/cve.py#L83) for CVEDataset, shouldn't they only be built on CVEs that have *no* `CPEConfiguration` records? Also, [self.cves_with_cpe_configurations](https://github.com/crocs-muni/sec-certs/blob/b5b0e9d1798338dca205b001e7b38f02c8929a89/src/sec_certs/dataset/cve.py#L62) essentially stores some CVEs that are already stored in `self.cves`, right? Shouldn't we delete them from `self.CVEs` then?
    - But we need to be careful about serialization. 
- Basically, I'm worried that we store some CVEs twice, and that we also run matching on all CVEs, and then on CVEs with `CPEConfiguration` records, which deteriorates the performance. This speaks into favour of having all CPEs stored in `CPEConfiguration` when they are part of `CVEDataset`. 
- Could you please doublecheck statistics in `vulnerabilities.ipynb` and check that they didn't change much? At least that's what I'd expected. Number of detected certs with CVEs should rise IMO, and those would be the certs that we newly match. 

I know that this PR takes time and that I always require some changes. It's that I'm not happy with the result just yet. If you'd prefer, I can finish the work so that you can move to something potentiallly more interesting. Let me know in any case. Thanks!

src/sec_certs/sample/cve.py
GeorgeFI: There is no any specific reason for storing `CPEConfiguration` as a list. I just wanted to keep the same data structure as the `CPE` records have.

adamjanovsky: Sure, that makes sense. Maybe  there's a reason why CPEs are a list, but I cannot recall any. Could you pls try to refactor both to sets and see what happens?

GeorgeFI: @adamjanovsky Refactoring seems to be okay - at least the pipeline of downloading CVEs, processing them and building the `CVEDataset` is passing. Due to the usage of `CVEs` in `pandas_columns` I will also investigate the Jupyter notebooks to check if the change does not break the analysis code.

adamjanovsky:
@GeorgeFI finalized my checks and we're good for merge 🎉 , thank you for your effort 👍.

Few thinks I've adjusted:
- I simplified parsing of nist dictionary when loading CVE dataset
- I simplified some fixtures and tests
- I reverted to using lists of CPEs instead of sets (complex reasons, I can describe in person if needed)
- I reverted to using CPEs instead of CPE uris in `CPEConfiguration` objects
- I discarded some auxillary CVE filtering (that you didn't introduce, but I've figured it's not helping with runtime or memory consumption).

adamjanovsky (approved):

PR #308, opened 2023-01-27
Replace linters with ruff

adamjanovsky:


adamjanovsky:
@J08nY could you pls take a look at Section 3 at https://github.com/charliermarsh/ruff#table-of-contents ? I'd consider introducing more checks, but some are very opinionated. I've went through the list and these are some that I propose:

- https://pypi.org/project/flake8-annotations/
- https://pypi.org/project/flake8-use-pathlib/
- https://pypi.org/project/flake8-simplify/
- https://pypi.org/project/flake8-return/
- https://github.com/charliermarsh/ruff#flake8-comprehensions-c4 this, among others, forbids lambdas, map and filter in favor of comprehensions. Something that I recently started doing, but I remember that you quite liked lambdas :D

Thoughts? These can be easily added via ruff.

J08nY:
> * https://pypi.org/project/flake8-annotations/

I wouldn't go for this. We don't have annotations in places where they are too hard or too messy and I don't want to be forced to handle that.

> * https://pypi.org/project/flake8-use-pathlib/

I mean I like pathlib but similar to the above, sometimes it does not make sense. I am not sure what this will do with our code. If you see the results and think they are sensible/fixable then maybe?

> * https://pypi.org/project/flake8-simplify/

Yeah why not, but make it not fail the build just tell me.

> * https://pypi.org/project/flake8-return/

I mean whatever.


> * https://github.com/charliermarsh/ruff#flake8-comprehensions-c4 this, among others, forbids lambdas, map and filter in favor of comprehensions. Something that I recently started doing, but I remember that you quite liked lambdas :D

I don't care for lambdas much, sometimes some things are more readable than others. Similarly than above, lets see what the results are on the code and whether we like the proposed changes.

adamjanovsky:
@J08nY 

> https://pypi.org/project/flake8-annotations/

this is really premature, never meant to implement it, don't know why I listed it 😆 

> https://pypi.org/project/flake8-use-pathlib/

[Resulted in just few suggestions](https://github.com/crocs-muni/sec-certs/pull/308/commits/94bd822d6defb26afb7c06c729bba707f92222be), gonna enforce it .

> https://pypi.org/project/flake8-return/

Many opinionated suggestions, implemented some but will not introduce this as a check.

> https://github.com/charliermarsh/ruff#flake8-comprehensions-c4

[simplified few things](https://github.com/crocs-muni/sec-certs/pull/308/commits/5f89fa0437f345ba8aa67c62aa68d262005a1e00), I'll keep it there.

> https://pypi.org/project/flake8-simplify/

[quite nice changes](https://github.com/crocs-muni/sec-certs/pull/308/commits/e5465f38f053941e0bcf3387da4e49394e169ee0), not opinionated. I don't thing I can set this not to fail. I started enforcing it. If we ever find some problematic instance or we feel it's more trouble than benefit, we can disable.

PR #318, opened 2023-02-22
Feat: add info retrieval split including labeling metadata

dmacko232:
Add information retrieval split that includes also data useful for labeling of images (page counts, page indices to label).


adamjanovsky (approved):
👍

PR #322, opened 2023-03-17
Reference analysis

### LINKED ISSUE #256: Detect re-certification
We could detect when a certificate is a re-certification as many such cases include the same wording, which in case of BSI is:
 - `This is a re-certification based on BSI-DSZ-CC-0417-2008.`
 - `a re-certification based on the certificate BSI-DSZ-CC-0171-2002`

Other schemes have similar wording in certification reports.

adamjanovsky:
This PR adds models to annotate references. 

Things to resolve:
- [x] This reverts from `billiard` to `multiprocessing`, check that we're ok with that
- [x] Label test examples for annotations -- otherwise the annotator will not work in production
- [x] Add analysis of references
- [x] If we are to further process the references, consider storing the labels in Enum. Right now they're just string
- [x] ~~Test that the annotator works in production.~~
- [x] Decide how to approach nlp dependencies. We either need to add new extra `[nlp]` and guards or keep them in ordinary requirements.
- [x] Maybe write some tests? But that's going to be painful
- [ ] ~~Actually call `annotate_references()` from somewhere. Probably from `analyze_certificates()`?~~

As a byproduct, this closes #256


adamjanovsky:
Just a quick comment on the current state of affairs:
- Soon, hyperparameter tunning will be done. Scripts for hyperparam tuning will not be versioned, I'll back them up separately at GDrive
- I don't think that the annotation engine is production-ready. For that reason, I kept the necessary files in `sec_certs.model.references_nlp` package that has some complex dependencies. Any import from that package requires `sec-certs` installation with `[nlp]` extra. A check is done on any import from that package.
- This requires that all `nlp` deps are also installed on dev machines -- otherwise, linters will likely complain
- In summary, this PR adds:
    - Backend for Reference annotations
    - Notebooks with reference annotation analysis
    - Versioned manual annotations, scripts to measure inter-annotator agreement

PR #326, opened 2023-03-29
switch to pydantic in settings management

adamjanovsky:
Management of global configuration variable is now left to Pydantic. This means:

- No json schema, this is replaced by Pydantic specification
- Description of default values and types is now accessible from docs, not from the example yaml
- No default `settings.yml` is needed anymore, the default values are stored in the class
- One can define new settings files just by overriding some keys, see test settings e.g. where only a single key is overriden.
- One can set keys with environment variables -- handy for secerts
- Some old keys got deleted, some sensible integer boundaries on existing keys were introduced
- I wrote some tests to check that new config works properly


J08nY (approved):
Looks good to me.

PR #329, opened 2023-04-11
Add CC scheme and FIPS IUT/MIP matching

### LINKED ISSUE #278: Enhance data collected from CC schemes and match to portal
For some schemes there is more information on the pages they provide that we do not collect currently:

- [x] [Australia](https://www.cyber.gov.au/acsc/view-all-content/programs/australian-information-security-evaluation-program): We can follow the link for products in evaluation and extract more stuff there.
- [x] [France](https://www.ssi.gouv.fr/en/products/certified-products/#cartes-a-puce): We can follow the link and extract more stuff. Also we can download the documents from there and compare against those from CC portal.
- [x] [Germany](https://www.bsi.bund.de/EN2021/Topics/Certification/certified_products/certified_products_node.html): We can follow the link and extract more stuff. Also we can download the documents from there and compare against those from CC portal.
- [x] [Japan](https://www.ipa.go.jp/security/jisec/jisec_e/certified_products/certfy_list_e31.html): We can follow the link and extract more stuff. Also we can download the documents from there and compare against those from CC portal. This can be also done for the in evaluation list.
- [x] [Norway](https://sertit.no/certified-products/category1919.html): We can follow the link and extract more stuff. Also we can download the documents from there and compare against those from CC portal. This can be also done for the archived list.
- [x] [Korea](https://itscc.kr/certprod/list.do?product_class=1): We can follow the link and extract more stuff. Also we can download the documents from there and compare against those from CC portal. This can be also done for the archived list.
- [x] [Sweden](https://www.fmv.se/verksamhet/ovrig-verksamhet/csec/certifikat-utgivna-av-csec/): We can follow the link and extract more stuff. Also we can download the documents from there and compare against those from CC portal. This can be also done for the archived and in evaluation lists.
- [x] [USA](https://www.niap-ccevs.org/Product/PCL.cfm): We can follow the link and extract more stuff. Also we can download the documents from there and compare against those from CC portal. *The product pages contain additional documents with details of the evaluation process that should have their own processing pipeline ideally*.
- [x] Netherlands
- [x] Singapore

Also, we should create a mapping between these collected entries and the main CC portal dataset entries. The matching could be done as follows:
 1. Try to match on canonicalized cert_ids (if the scheme page has them, maybe we can also extract them from the documents if the scheme has them, atleast the filenames).
 2. Maybe try to match on document hash if the scheme has them.
 3. Match on product name + category + vendor?
 4. Give up matching?

After we have this we should evaluate how good the matching is.

### LINKED ISSUE #293: Track module movement through IUT, MIP and cert list
A crypto module in FIPS goes through several stages (and substages) when it is present in various lists (IUT, MIP lists) and then finally a cert is issued and it is present in the final certificate list. We could track (or rather link together) how a particular module moves through these lists. The lists should have a shared key of `module_name + vendor_name` though it may not be a perfect mapping so maybe some fuzziness/special-casing might be required. We could then also give stats on the IUT/MIP parts of a certificate lifetime on the certificate page (e.g., this certificate spent 30 days in the "In-review"  stage of the MIP list). This would also allow us to get correlations between the final certificate properties (e.g. vendor, crypto used, ...) and how long it spent in the various IUT/MIP stages.

J08nY:
Will eventually fix #293, fix #278.

- [x] Add `match_snapshot` method to FIPS to process a whole snapshot.
- [x] Add `match_all` to process a set of CC scheme entries.
- [x] Fix rapidfuzz use.
- [x] Move the CPE title/vendor canonicalization functions to shared utils.
- [x] Use the title/vendor canonicalization functions to improve performance.
- [x] Add tracking of MIPEntries within a MIPDataset (across snapshots)
- [ ] Add mapping between IUT and MIP.
- [x] Actually add this processing to the CC and FIPS datasets and enrich them with the data.
- [x] Enhance the CC scheme data for #278.


J08nY:
All green :heavy_check_mark: les gooo!!

PR #345, opened 2023-06-16
Metadata Bindings

xmoravec:
This PR implements the metadata layer for sec-certs, including the pipeline processing the binding files, creating binding records in a new MongoDB collection, and the web presentation of the data.

### DIFF
--- a/sec_certs_page/common/bindings.py
+++ b/sec_certs_page/common/bindings.py
@@ -1,0 +1,529 @@
+import logging
+import os
+import json
+import datetime
+import asyncio
+import hashlib
+import shutil
+import base64
+from urllib.parse import urlparse, unquote
+import urllib
+from typing import Tuple, Union, Any
+import requests
+import jwt
+import click
+from flask import current_app
+from pymongo.collection import Collection
+
+
+from .. import mongo
+
+logger = logging.getLogger(__name__)
+Document = Any
+
+class Binding():
+    """ A binding data class
+    """
+
+    def __init__(self, cert_id: str, header_name: str):
+        """Binding object constructor
+
+        Args:
+            cert_id (str): certificate id
+            header_name (str): header name
+        """
+        self.cert_id: Union[str, int] = cert_id if not cert_id.isdigit() else int(
+            cert_id)
+        self.header_name: str = header_name
+        self.header_data: list[object] = []
+
+    def match_name(self, name: str) -> bool:
+        """ Function to match a string to the header_name
+
+        Args:
+            name (str): Match candidate
+
+        Returns:
+            bool: Whether the name matches to the binding obj header_name
+        """
+        return name == urllib.parse.quote(
+            self.header_name) or name == self.header_name
+
+    def update_headers_if_match(self, name: str, header: list[object]) -> None:
+        """ Update header_data if this objects header_name matches with given name
+
+        Args:
+            name (str): Match candidate
+            header (list[object]): New header data
+        """
+        if self.match_name(name):
+            self.header_data = header
+            for h in self.header_data:
+                if "metadata_preview" in h and verify_url(h["metadata_preview"]):
+                    click.echo(f"Converting image to base64")
+                    h["metadata_preview"] = get_base64_preview(h["metadata_preview"])
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+    def __repr__(self) -> str:
+        return f"Binding {self.cert_id} to {self.header_name}"
+
+    def __getitem__(self, key: str) -> Union[str, int, list[object]]:
+        return {
+            "cert_id": self.cert_id,
+            "header_name": self.header_name,
+            "header_data": self.header_data
+        }[key]
+
+def get_base64_preview(url: str) -> str:
+    """ Function to get the base64 encoded representation of an image at a given URL
+
+    Args:
+        url (str): Preview pic location
+
+    Returns:
+        str: Base64 encoded preview image
+    """    
+    response = urllib.request.urlopen(url)
+    file_content = response.read()
+    base64_content = base64.b64encode(file_content)
+    base64_string = base64_content.decode('utf-8')
+    return base64_string
+
+def url_to_local_path(url: str) -> str:
+    """ Transorms local URL to local path
+
+    Args:
+        url (str): URL to transform
+
+    Returns:
+        str: Transformed URL
+    """
+    parsed_url = urlparse(url)
+    path = unquote(parsed_url.path.strip("/"))
+    return path
+
+
+def github_url_to_api(url: str) -> str:
+    """ Transform the standard public github URL to a github API URL
+
+    Args:
+        url (str): URL to transform
+
+    Returns:
+        str: Transformed API URL
+    """
+    parsed_url = urlparse(url)
+    api_url = f"https://api.github.com/repos{parsed_url.path}"
+    if parsed_url.query:
+        api_url += f"?{parsed_url.query}"
+    return api_url
+
+
+def get_filename_from_url(url: str) -> str:
+    """ Parse out the filename from a given url
+
+    Args:
+        url (str): URL to parse
+
+    Returns:
+        str: Parsed out filename
+    """
+    path_components = url.split("/")
+    filename = path_components[-1]
+    if not filename:
+        filename = path_components[-2]
+    return os.path.basename(filename)
+
+
+def verify_timestamp(timestamp: str) -> bool:
+    """ Verifies timestamp validity
+
+    Args:
+        timestamp (str): timestamp to validate
+
+    Returns:
+        bool: Whether the given timestamp is valid
+    """
+    try:
+        datetime.datetime.fromisoformat(timestamp)
+        return True
+    except ValueError:
+        return False
+
+
+def verify_url(url: str) -> bool:
+    """ Verifies URL validity
+
+    Args:
+        url (str): URL to validate
+
+    Returns:
+        bool: Whether the given URL is valid
+    """
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+    except ValueError:
+        return False
+
+
+def verify_jwt(data: dict[str, Union[str, object]]) -> bool:
+    """ Verifies the given object jwt validity
+
+    Args:
+        data (dict[str, Union[str, object]]): object to validate
+
+    Returns:
+        bool: Whether the JWT in the data object is valid
+    """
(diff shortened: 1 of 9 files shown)


J08nY:
@xmoravec Should I look at this and give it a review? Is it ready?

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+    """ Transorms local URL to local path
+
+    Args:
+        url (str): URL to transform
+
+    Returns:
+        str: Transformed URL
+    """
+    parsed_url = urllib.parse.urlparse(url)
+    path = urllib.parse.unquote(parsed_url.path.strip("/"))
```
J08nY, line 83: This also strips the trailing `/`. Should it?
Will this be used with world-provided/untrusted URLs?
What about directory traversal? I think we should be vary.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+def get_filename_from_url(url: str) -> str:
+    """ Parse out the filename from a given url
+
+    Args:
+        url (str): URL to parse
+
+    Returns:
+        str: Parsed out filename
+    """
+    path_components = url.split("/")
```
J08nY, line 112: What about using `urlparse` here and only splitting its path?

```suggestion
    path_components = urllib.parse.urlparse(url).path.split("/")
```

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+    filename = path_components[-1]
+    if not filename:
+        filename = path_components[-2]
+    return os.path.basename(filename)
```
J08nY, line 116: Why is basename necessary? The filename should have no `/` at that point so basename will just be the identity function. This whole function is quite weird by looking at the last path component and then the second last, which will only happen in case the path ends in a trailing slash. I think the function should not be so benevolent.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+        data (dict[str, Union[str, object]]): object to validate
+
+    Returns:
+        bool: Whether the JWT in the data object is valid
+    """
+    jwt_token = data['JWT']
+    del data["JWT"]
+    encoded_jwt = jwt.encode(
+        data, key=current_app.config["BINDINGS_PUBLIC_KEY"], algorithm="RS256")
+    if not jwt_token:
```
J08nY, line 164: This check can happen earlier.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+    """ Verifies the given object jwt validity
+
+    Args:
+        data (dict[str, Union[str, object]]): object to validate
+
+    Returns:
+        bool: Whether the JWT in the data object is valid
+    """
+    jwt_token = data['JWT']
+    del data["JWT"]
```
J08nY, line 161: Modifying data here is nasty. Python dicts are mutable, make a local copy to verify inside the function.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+    encoded_jwt = jwt.encode(
+        data, key=current_app.config["BINDINGS_PUBLIC_KEY"], algorithm="RS256")
+    if not jwt_token:
+        click.echo("JWT token is missing")
```
J08nY, line 165: Click should only be used for CLI-called stuff. Will this only ever be called using some CLI command or automated? If the latter, this should either be a logging call or an exception thrown.

sec_certs_page/templates/cc/entry/index.html.jinja2
```diff
@@ -446,6 +446,11 @@
                 {% endif %}
             </div>
 
+            {% if additional_metadata %}
+                <h3 class="mt-3">Additional Metadata</h3>
+                {{ metadata_binding_card("cc", additional_metadata["metadata_headers"]) }}
```
J08nY, line 451: The first argument `doc_prefix` is not meant to be "cc" or "fips" but either `"cert"` or `"target"` based on whether the card is inside the "certification report" part of the page or the "security target" part. In this case the card is part of neither, so the doc prefix need not be used, instead something like "metadata" should be used to make the element ids unique.

sec_certs_page/templates/fips/entry/index.html.jinja2
```diff
@@ -336,6 +336,11 @@
                 {% endif %}
             </div>
 
+            {% if additional_metadata %}
+                <h3 class="mt-3">Additional Metadata</h3>
+                {{ metadata_binding_card("fips", additional_metadata["metadata_headers"]) }}
```
J08nY, line 341: Same as in CC, ditch the doc_prefix.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+        purge (bool, optional): Whether to purge the data before update. Defaults to True.
+        verbose (bool, optional): Whether to click.echo out extra information. Defaults to False.
+    """
+    bindings = process_and_download_all(
+        download_bindings,
+        download_headers,
+        download_metadata,
+        verbose=verbose)
+    click.echo(
+        "\033[32m" +
```
J08nY, line 502: click can do colors natively, use that.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+        process_header_files(
+            headers_path,
+            metadata_path,
+            bindings,
+            download_metadata,
+            verbose=verbose))
+    return bindings
+
+
+def purge_headers_data(collection: Collection[Document]) -> None:
```
J08nY, line 444: Unnecessary one liner, just do this at the call site.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+        file_url (str): URL of the file to download
+        file_path (str): Path to where to download the file
+        sha256 (str, optional): Expected hash of the downloaded file. Defaults to "".
+        verbose (bool, optional): Whether to click.echo out extra information. Defaults to False.
+    """
+    if verbose:
+        click.echo(f"Downloading {file_url}...")
+    response = requests.get(file_url, timeout=1000)
+    if sha256:
+        file_sha256 = hashlib.sha256(response.content).digest().hex()
```
J08nY, line 185:
```suggestion
        file_sha256 = hashlib.sha256(response.content).hexdigest()
```

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+        if file_sha256 != sha256:
+            click.echo(
```
J08nY, line 187: This should abort right? Why is there verification at all if it still saves the file and continues?

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+    if verbose:
+        click.echo(f"Downloading bindings files from {url}")
+
+    if url.startswith("https://github.com"):
+        tasks = []
+        repo_url = url.split("/tree")[0]
+        dir_path = url.split("/main")[1]
+        api_url = github_url_to_api(repo_url) + "/contents" + dir_path
+
+        response = requests.get(api_url)
```
J08nY, line 216: We really need to be authenticated to Github in order for this to work, the rate limit is quite steep. Some 60 requests per few hours or so. I think our batch processing, even if only done daily or weekly could go over that with enough bindings.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+            filepath = os.path.join(bindings_dir, filename)
+            with open(filepath, "r", encoding="utf8") as f:
+                data = json.load(f)
+                if not verify_jwt(data):
+                    click.echo(
+                        f"JWT verification failed processing binding file {filename}")
+                    continue
+                for item in data["data"]:
+                    timestamp = item["timestamp"]
+                    if not verify_timestamp(timestamp):
```
J08nY, line 314: Why is this necessary? Either the JWT verification passed, in which case the data is signed (by whoever, I don't know the full system) and we trust it and therefore do not need this secondary verification which is quite weak. Or is this some data sanitization to make sure the format is right?

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+                        urllib.parse.unquote(url))
+                    if not verify_url(url):
+                        click.echo(
+                            f"Invalid metadata_header_url for file: {url}")
+                        continue
+                    if not download_headers:
+                        continue
+
+                    for cert_id in item["certificate_ids"]:
+                        bindings.append(Binding(cert_id, header_file_name))
```
J08nY, line 328: These lines should be flipped right? the check should come after the items are added to the bindings list.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+            with open(file_path, "r", encoding="utf8") as f:
+                if verbose:
+                    click.echo(f"Processing {file_name}")
+                json_obj = json.load(f)
+                if not verify_jwt(json_obj):
+                    click.echo(
+                        f"JWT verification failed for header file {file_name}")
+                    continue
+                for data_obj in json_obj["data"]:
+                    if not verify_timestamp(data_obj["timestamp"]):
```
J08nY, line 374: Same, not sure why we check this.

sec_certs_page/common/bindings.py
```diff
@@ -0,0 +1,510 @@
...
+        binding: Binding,
+        verbose: bool = False) -> None:
+    """ Updates a record in mongoDB with new header_data
+
+    Args:
+        collection (Collection[Document]): MongoDB collection to update the records in
+        binding (Binding): An object representing the binding - cert_id and header_data
+        verbose (bool, optional): Whether to click.echo out extra information. Defaults to False.
+    """
+    cert_id = binding["cert_id"] if "NIST" not in binding["cert_id"] else int(
```
J08nY, line 465: This is nasty. The binding should just have another attribute that says whether it is CC or FIPS. Nothing like "NIST-" prefixed cert IDs exists anywhere.

J08nY (changes requested):
Several things need to be handled.

petrs:
@xmoravec Hi Eric, do you find the comments from @J08nY actionable and are you willing to incorporate before merging? It would be better to have things merged otherwise this PR may become incompatible soon due to other changes. Thank you for consideration

J08nY:
Let's close this.

PR #348, opened 2023-07-13
don't clone in dockerfile, copy local repo contents

### LINKED ISSUE #339: Don't clone in Dockerfile, copy contents
**Describe the bug**
The dockerfile now clones the project from GitHub. Instead, we should copy the contents locally. Currently, it's impossible to install the version from any place but origin.

adamjanovsky:
This closes #339

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+# The directory with the datasets
+dataset/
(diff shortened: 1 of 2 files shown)


adamjanovsky:
@J08nY This was something that you complained about from time to time. Should be fixed now.

J08nY:
This should also be implemented in the dockerfile on the page branch, so that the web also uses this approach for both the main repo and the page.

PR #349, opened 2023-07-14
Dockerfile: install from local resources

### LINKED ISSUE #339: Don't clone in Dockerfile, copy contents

adamjanovsky:
@J08nY this should close #339 for page branch.

It's not really elegant because we cannot reuse the Dockerfile in the main branch which leads to code duplication. 

Also, one drawback that I discovered is that the checkouting of the branches performed here requires that the changes in your local repository instance must be commited (otherwise the checkout will not perform). 

Could you pls check that the page works as expected?

I got this error

```
(venv) ~/phd/projects/certificates/sec-certs  $ docker run page
Traceback (most recent call last):
  File "/home/user/venv/bin/flask", line 8, in <module>
    sys.exit(main())
  File "/home/user/venv/lib/python3.10/site-packages/flask/cli.py", line 1050, in main
    cli.main()
  File "/home/user/venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/user/venv/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/user/venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/user/venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/user/venv/lib/python3.10/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/user/venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/user/venv/lib/python3.10/site-packages/flask/cli.py", line 911, in run_command
    raise e from None
  File "/home/user/venv/lib/python3.10/site-packages/flask/cli.py", line 897, in run_command
    app = info.load_app()
  File "/home/user/venv/lib/python3.10/site-packages/flask/cli.py", line 308, in load_app
    app = locate_app(import_name, name)
  File "/home/user/venv/lib/python3.10/site-packages/flask/cli.py", line 218, in locate_app
    __import__(module_name)
  File "/home/user/sec-certs-page/sec_certs_page/__init__.py", line 48, in <module>
    if not app.testing and app.config["SENTRY_INGEST"]:  # pragma: no cover
KeyError: 'SENTRY_INGEST'
```

But I don't see any manual on how to invoke the docker image, so it may be my problem. 

If you find this ok, feel free to merge.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+# The directory with the datasets
+dataset/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,72 @@
-FROM seccerts/sec-certs
+FROM ubuntu:jammy-20220428 AS intermediate
+
+RUN apt-get update
+RUN apt-get install git -y
+
+WORKDIR /tmp
+COPY . /tmp
+RUN mkdir /output
+RUN git ls-files | xargs cp -r --parents -t /output
+COPY .git /output/.git
+
+FROM ubuntu:jammy-20220428
+
+ENV USER="user"
+ENV NB_UID=1000
+ENV NB_GID=1000
+ENV HOME /home/${USER}
+
+#installing dependencies
+RUN apt-get update
+RUN apt-get install python3 -y
+RUN apt-get install python3-pip -y
+RUN apt-get install python3-venv -y
+RUN apt-get install git -y
+RUN apt-get install curl -y
+
+# Install dependencies fo PyPDF2 and pdftotext
+RUN DEBIAN_FRONTEND="noninteractive" apt-get -y install tzdata
+RUN apt-get install build-essential libpoppler-cpp-dev pkg-config python3-dev -y
+RUN apt-get install libqpdf-dev -y
+RUN apt-get install default-jdk -y
+RUN apt-get install tesseract-ocr tesseract-ocr-eng tesseract-ocr-deu tesseract-ocr-fra -y
+
+
+RUN groupadd -g ${NB_GID} -o ${USER}
+RUN adduser --disabled-password \
+  --gecos "Default user" \
+  --uid ${NB_UID} \
+  --gid ${NB_GID} \
+  ${USER}
+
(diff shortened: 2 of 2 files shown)


J08nY:
I don't understand why this cannot use the sec-certs docker image and build on top of it (as it was before) but just get the "page" branch contents using the new approach. That makes sense right? Or does it not? I am confused to what the various approaches bring or require.

adamjanovsky:
> I don't understand why this cannot use the sec-certs docker image and build on top of it (as it was before) but just get the "page" branch contents using the new approach. That makes sense right? Or does it not? I am confused to what the various approaches bring or require.

IMO that would pull whatever version of sec-certs just sits in the dockerhub, unless I am mistaken. 

The approach that I proposed checkouts at HEAD of your local repository instance. Possibly, we could add an argument to allow the user to specify the branch from which checkout shall be made. This may be handy in situations when you're developing something and want to check that page works. 

I don't really care about the page. If we want it to get the `sec-certs` code from DockerHub release, we can certainly do it and simplify the Dockerfile.

J08nY:
> IMO that would pull whatever version of sec-certs just sits in the dockerhub, unless I am mistaken.

Yes and I thought that would be a good idea right? I am thinking "reuse = good" here, but I may be wrong.

With the setup as this PR adds I worry about repository clones that do not have the main branch. I am doing the sec-certs-page development in a local clone that does not have the main tool branch at all. So this may not work at all.

I think pointing at the sec-certs docker image from the docker hub (or maybe built locally by the user?) makes the most sense.

PR #351, opened 2023-07-22
Analysis of FIPS references

GeorgeFI:
This is initial phase of the FIPS reference analysis. In some cases I tried to mirror the CC reference analysis. I also added some extra cells to provide more insights to FIPS references data. I open this PR as draft, because I think I can get much more from the data.


GeorgeFI:
@adamjanovsky Automatic part of the FIPS analysis is finished from my side and ready for final review.

adamjanovsky (approved):
Hey, thanks for your effort on this! :+1: I like the overall structure and the plots. Some nitpicks are below:

## Code

- `Distribution of total number of references per cert since 2016` plot could also list certificates with 0 references, this would give a better overall view IMO
- `Distribution of levels with assigned at least one CVE` this plot can be made more informative if the columns are turned relative, i.e., normalized by the overall number of certificates with the given level. Sample applies for the two plots below (CVEs per category and status)
- You mention the certificate with dgst `10acb87b3040f4e1 ` as one with CVEs and being referenced by 7 other certs. Interestingly, this one is from year 2020 and is already historical. Could you please manually investigate why is it already historical? Is it because the CVE?
- `Temporal evolution of references in time` this could get a more informative title. Is it a temporal evolution of the number of certificates that reference other certificate?
- There are two plots called `Histogram of CVEs`. The second plot is not given a title and its meaning is unclear without studying the code. Please add an informative title. 
- Drawing of the network graphs (e.g., `Certificate overview`) could be wrapped into functions. Please define functions, they may be recycled for further use. Ideally also pack the functions outside of the jupyter notebook into some appropriate place in `utils` module. 

## Write-up

- I like the write-up at Trello :+1:. 
- Some of the markdown images are not rendered for me, could you please investigate? (`![[time_analysis_mean_references.png]]`)
- Please adjust the finding w.r.t. my feedback to the code above

Otherwise, I'm aproving right away, so that when you implement the changes, feel free to merge.

GeorgeFI:
Thank you for the review. I will implement the changes by the end of this weekend :)

PR #355, opened 2023-08-24
anonymize the docs

adamjanovsky:


adamjanovsky:
@J08nY I've anonymized the docs. The problem is that they link the markdown files from the repository, so I needed to anonymize them as well. I just dropped the mention of authors from the GitHub readme. I suppose we can live with that for two months. People will still see the maintainers and can open an issue or mail the maintainers directly. 

Also, it's not perfectly anonymous, some badges e.g., lead to PyPi with authors, etc. But we likely don't need to go that far.

J08nY:
Agreed.

PR #367, opened 2023-10-18
Evaluate possible switch from pdftotext to PyMuPDF

### LINKED ISSUE #364: Evaluate possible switch from pdftotext to PyMuPDF
We should evaluate whether it's worth switching from `pdftotext` to [PyMuPDF](https://pymupdf.readthedocs.io/en/latest/) for data extraction. We should:

1. @dmacko232 provide code that substitutes [pdftotext conversion](https://github.com/crocs-muni/sec-certs/blob/f59e517490112c8baeb56c81a5869774ad1b821b/src/sec_certs/utils/pdf.py#L73) with an identical functionality achieved using PyMuPDF
2. @adamjanovsky run the new pipeline on full dataset and provide results to @J08nY 
3. @J08nY compare the outputs on both datasets, decide if we should switch

Note: @adamjanovsky will handle the second step instead of originally advertised @GeorgeFI

adamjanovsky:
This closes #364

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@
     "networkx",
     "pydantic",
     "psutil",
+    "pytesseract",
   ]
 
   [project.optional-dependencies]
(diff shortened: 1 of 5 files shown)


adamjanovsky:
Guys sorry, I messed up the branches. Cannot adjust this, creating new PR.

PR #368, opened 2023-10-18
Feat/pymupdf experiment

### LINKED ISSUE #364: Evaluate possible switch from pdftotext to PyMuPDF

adamjanovsky:
This closes #364


adamjanovsky:
@J08nY The datasets computed with `pdftotext` and `pymupdf` are available at Aura, sitting at `/var/tmp/xjanovsk/certs/sec-certs/dataset/toy_dataset_100_certs`; you should have read access there. 

Performance wise, the processing speed seems similar.

adamjanovsky:
@dmacko232 Do we know what are the internal dependencies of `pymupdf` package? Could we drop dependency on `poppler` if we make a switch? 

Also, we're scanning some tables in FIPS documents with some java tool. Could we get rid of the java dependency as well?

dmacko232:
@adamjanovsky Poppler is not dependency. The java thing should not be dependency either I guess in case of pymupdf.

adamjanovsky:
Closing this. The details of what would it take to get pymupdf surpass pdftotext in terms of output quality are described in #364

PR #370, opened 2023-10-22
feat: Initial implementation of dataset versioning

GeorgeFI:


adamjanovsky:
I briefly reviewed this and it looks OK. I guess tests need to change as well. We should also add a new test and check if the warning was issued. I'd like to add two types of warnings:

- Just some note in case the release matches and there's some post-version mismatch
- Real warning when the release of the dataset is older than the release of the tool (or vice versa).

adamjanovsky (changes requested):
I briefly reviewed this and it looks OK. I guess tests need to change as well. We should also add a new test and check if the warning was issued. I'd like to add two types of warnings:

- Just some note in case the release matches and there's some post-version mismatch
- Real warning when the release of the dataset is older than the release of the tool (or vice versa).

J08nY:
Hey all, what is the status of this? Seems quite stale? A rebase would help.

J08nY:
Pinging this, @GeorgeFI  @adamjanovsky. Can we get this resolved?

adamjanovsky:
Not really, no progress there. Unassigning @GeorgeFI as he won't work on this any further.

J08nY:
Lets either merge or close then. It does not make any sense to keep these three line change here.

Closing.

PR #386, opened 2024-02-02
Cleanup cert-id rules

J08nY:
The cert-id rules work, but are a mess. This PR will clean them up for presentation in the paper.

Before this, the regular cert_id regexes were used to extract
the cert_id from the report filename. However, the filenames often
do not use the same cert_id format, but contain all of the information
necessary to reconstruct the cert_id, but with different order for example.

This commit along with those before it introduce a new set of regular
expressions that better match the ones in the filenames. To extract
the correctly formatted canonical cert_id, the regexes are used to
obtain the parts of the cert_id (using named groups in regexes) and
those are then reconstructed into a canonical version of the cert_id
via one of the scheme-dependent functions.


J08nY:
Now that this is merged and I did a full run I will use this PR for some insights/issues.
 - CA rule causes many false positives like `1 1 1` or `2021-02-13`.
 - Some US ids are missed due to bad ? placement in regex.
 - The number of certificates without ID increased a bit, but I think that is due to us being overly eager before and actually matching stuff we should not have. Though a proper evaluation should be done.

PR #403, opened 2024-04-25
replace seccerts -> sec-certs

adamjanovsky:
**Caveats**:
- Dockerhub account [cannot be renamed](https://github.com/docker/roadmap/issues/44) now. Creating a new one, we would need to keep track of historical images. I suggest keeping the account `seccerts`.
- `sec_certs` is preferred as a package name over `sec-certs` as is common in Python projects.
- @J08nY could you please merge this once https://sec-certs.org is up and running?


J08nY:
I still feel like this is such a pointless change 😀.

PR #424, opened 2024-07-19
New digests

### LINKED ISSUE #420: CCportal URL change
Common Criteria portal recently moved from artifact urls of the form:
```
https://www.commoncriteriaportal.org/files/epfiles/2019-44-INF-3227.pdf
```
to the form
```
https://www.commoncriteriaportal.org/nfs/ccpfiles/files/epfiles/2019-44-INF-3227.pdf
```

Subsequently, this broke our certificate digest, as it was computed as a a hash of some metadata that included the `cert_link`.
My fix in https://github.com/crocs-muni/sec-certs/commit/b333fd24dcbd1e4aa08584343cfb5b9784364a03 was to remove the new part in the URL, as it seemed the old links were still valid. This fixed the problem and did not cause havoc on the database on the server, which would mean we would lose history of the certificates.

However, it turns out that new certificates have links of the new form and the old form links *do not work* for their artifacts.
Example being:
(works) https://www.commoncriteriaportal.org/nfs/ccpfiles/files/epfiles/NSCIB-CC-2300182-01-CR2.pdf
(doesn't work) https://www.commoncriteriaportal.org/files/epfiles/NSCIB-CC-2300182-01-CR2.pdf

I do not have time to fix this at the moment, nor do I have a good idea for a fix.

Maybe we could keep the URLs as-is in the metadata but turn them into the old format for the digest computation. I do not think there will be conflicts between new and old format.

J08nY:
Fixes #420.


J08nY:
@adamjanovsky What do you think about this solution? It moves to what we discussed in #420, using the filename parts of the st and report links. It also keeps the old digest attributes, perhaps a better naming there would be possible. I was thinking of having "alt_dgsts" or "old_dgsts" property that would have the older digests and would be a set. What do you think? This is necessary to be able to map the old to the new on the site (as well as for other of our users for some kind of compat).

adamjanovsky:
Well, late to the review I guess :)

Anyways, I agreed with the proposal, didn't go through the implementation in the detail. One think to keep in mind is our usage of digests in Notebooks. I suppose that the backward-compatability redirect only applies to web, right? 

What's going to happen when you load an old dataset with a new tool version?

J08nY:
> Anyways, I agreed with the proposal, didn't go through the implementation in the detail. One think to keep in mind is our usage of digests in Notebooks. I suppose that the backward-compatability redirect only applies to web, right?

Yes right now. However, you can get the old digests from the certs.


> What's going to happen when you load an old dataset with a new tool version?

Interesting thought, haven't considered it much as I was looking at it from the POV of the web where a fresh run is running. However, I tested this a bit and what happens is that the new digests take precedence. So after loading it you get a dataset with new digests only (because `dgst` is a property computed from attributes) and if you store it again you will get the new digests also in the JSON. Basically, the old digests are ignored.

adamjanovsky:
> Interesting thought, haven't considered it much as I was looking at it from the POV of the web where a fresh run is running. However, I tested this a bit and what happens is that the new digests take precedence. So after loading it you get a dataset with new digests only (because dgst is a property computed from attributes) and if you store it again you will get the new digests also in the JSON. Basically, the old digests are ignored.

Cool, that's what I anticipated.

> Also, regarding the quick merge. I wanted to get this in quickly because of the Monday update run. The web has been broken for a while and I wanted to fix it. This was a pre-requisite.

Understood :+1:

PR #436, opened 2024-09-19
Fix NVD api download.

J08nY:
They changed the parameter names:
https://nvd.nist.gov/developers/vulnerabilities#:~:text=lastModStartDate%20%26%20lastModEndDate


J08nY:
Only CC sheme tests fail now. This can be merged.

PR #440, opened 2024-09-28
WIP: awesome curated list of sec-certs string searches

petrs:
The goal is to provide curated list of relevant and helpful searches over Common Criteria and FIPS140 document executed via sec-certs.org webpage.

### DIFF
--- a/awesome_search_examples.md
+++ b/awesome_search_examples.md
@@ -1,0 +1,79 @@
+# Awesome curated list of 'sec-certs' searches
+The goal is to provide a curated catalog of search strings over Common Criteria and FIPS140 certification artifacts executed on the https://sec-certs.org webpage.
+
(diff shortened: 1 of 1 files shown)


adamjanovsky:
Could we move this from root directory to docs folder? This way, we could display it via sphinx.

petrs:
> Could we move this from root directory to docs folder? This way, we could display it via sphinx.

sure, will do

petrs:
@adamjanovsky updated, is it now what you meant?

adamjanovsky:
> @adamjanovsky updated, is it now what you meant?

Yep 👍

J08nY:
Can we get this merged before the conference?

mukrop:
Is there anything necessary to do before the merge? What is the blocker? (I don't see any.)

Though, I would suggest linking this also from some place on the web (I don't think people would open documentation much). Can we, for example, link this from the search pages directly (as more advanced examples)?

PR #444, opened 2024-10-16
Fix mapping of certs to PPs broken due to CC URL change.

J08nY:
The PP links in the CSV/HTML are now slightly different than they were before. Since the PP dataset is static, it still has the old links. We use the links to match the PPs to the certs, hence that matching was broken with the change. This PR tries to fix the issue by mapping links to filenames before the matching is done.


adamjanovsky (approved):
That looks flawless, thanks 👍.

PR #446, opened 2024-10-17
Add a way to download full dataset archive (including PDFs) from the web.

J08nY:
As it says in the title. This is useful if someone wants to have the full dataset.
I think a utility method that would download dataset + auxiliary datasets while not downloading all the PDFs would also be quite helpful and would actually be what 90% people want.

Wdyt @adamjanovsky ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,8 @@ df_2015_and_newer = df.loc[df.year_from > 2014]
 df.year_from.value_counts().sort_index().plot.line()
 ```
 
-<!-- ## Authors
+## Authors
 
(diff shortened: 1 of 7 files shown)


adamjanovsky:
That would be a nice addition. Did you consider not adding new method? Instead we could add `download_artifacts: bool = False` optional argument into `from_web_latest()` method. What do you think?

J08nY:
Yes I did consider it, but right now that method downloads a "pathless" dataset just into memory, while this one needs a path(directory) where to save it. So it would need two arguments, which was a bigger change but we can go that route, we just need to figure out a sensible behavior for the combinations of arguments.

PR #451, opened 2024-10-31
feat: Add static charts for cc analysis

xvalec01:


J08nY:
Could we make these SVGs?

adamjanovsky:
@xvalec01, @petrs please find some of my FIPS plots attached
[plots.zip](https://github.com/user-attachments/files/17584986/plots.zip). They're in PDF format, but vectorized AFAIK.

xvalec01:
I made the images SVGs.

PR #453, opened 2024-10-31
Feat page index adjustment

mukrop:
Proposition of textual changes on the title page: Lead paragraph and feature list.

* The general idea stands on the messaging we created for the ICCC marketing materials.
* The wording in the individual feature sections is heavily based on the original wording.
* The collapsibles in the feature blocks were dropped as multiple people found it counter-intuitive that the opened text would contain important links.
* It's not ideal, but I believe an improvement over the current state.

Unfortunately, I don't have currently set up the local development environment, so cannot guarantee it will compile flawlessly (apologies!).

### DIFF
--- a/sec_certs_page/templates/index.html.jinja2
+++ b/sec_certs_page/templates/index.html.jinja2
@@ -5,13 +5,17 @@
         <div class="col-12 col-sm-10 mx-auto p-3 py-md-5">
             <div class="row mb-3 g-4 justify-content-between row-cols-1 row-cols-lg-2">
                 <div class="col-12 col-sm-5 order-1 order-sm-0">
-                    <h1>Security Certifications</h1>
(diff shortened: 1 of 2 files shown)


J08nY:
Looks good, though the texts for the vulnerability and notifications features are quite a bit longer than the other ones, giving it an unbalanced look. I will shorten them (the notifications mainly) and add appropriate small buttons for the vuln one to go to CVE/CPE search.

image

mukrop:
You're right, thanks. Adjust as you see fit.

J08nY:
Made it nicer.

image

J08nY:
Huh, this somehow broke the logo display at the index page.
Not sure what happened but will fix it.

image

J08nY:
Ok fixed.

Issue #464, opened 2025-01-13
Introduce `DocumentState` for `FIPSCertificate` class


adamjanovsky:
The `DocumentState` class is stored for various documents associated with `CCCertificate` class. The same class should be used to hold data in `FIPSCertificate` instances

J08nY:
Connected to this, there is quite a bit of deduplication possible. See what I have done in 5e983efc922a2a353332ffe75018547539ba03e4. Essentially, all of the boilerplate functions on the certificate objects that get parallelized: download_pdf, convert_pdf, extract_pdf_meta, extract_keywords are essentially the same for CC and PP and the only thing that changes is the: link, `DocumentState` object they store the results to and `PdfData` object they store the filename to. With FIPS unifying to use `DocumentState` we could simplify this greatly.

adamjanovsky:
Yes, I've seen that while copying the new functionality for PPs straight from the CC implementation. I didn't want to handle the abstraction at that time. Essentially, you just want to specify the action (download, convert, extract), the source attribute and the target attribute. Then you just execute.

J08nY:
The thing is that apart from DocumentState the methods set the "*_filename" attribute of PdfData. So basically there needs to be consistency between those two classes, and there is nothing enforcing it (be it typing or something at runtime) if we move to the `getattr/setattr` implementation.

adamjanovsky:
Here's my take on the code unification. 

Currently, document management is fragmented: local paths and processing statuses are tracked in DocumentState or scheme-specific InternalState classes, while extracted content (like metadata and keywords) is stored separately in PdfData structs. Also, while the conversion pipelines are nearly identical for all documents (Pdf get downloaded, converted, their keywords and metadata extracted), the implementation is duplicated across different certification schemes. 

I think a refactor is possible with a scheme-agnostic `Document` class that encapsulates everything related to a specific file (URLs, hashes, local paths, download/conversion state, metadata, and keywords). Certificates could store their artifacts in a standardized Documents container accessible via a unified DocType enum (e.g., `cert.documents[DocType.report]` or just `cert[DocType.report]`), eliminating redundant boilerplate and allowing us to share universal pipeline functions across the entire codebase. Instead of:

- `convert_reports()`
- `converts_security_targets()`
- `convert_certificates()`
- `convert_policies()`

we could have just `convert_document()`.

[...]

PR #466, opened 2025-01-23
Full fledged processing of protection profiles

### LINKED ISSUE #72: New API: Allow for full-fledged processing of protection profiles
Currently, the static dataset of protection profiles is loaded from [web storage](https://is.muni.cz/go/yodfo8). This dataset was created using [old API](https://github.com/crocs-muni/sec-certs/blob/6610a1ea0144d7dce72e713153df5d989668b6a4/process_certificates.py#L137).

The new tool should allow for full processing of protection profiles without downloading the static dataset.

adamjanovsky:
Closes #72


src/sec_certs/dataset/cc.py
```diff
@@ -164,38 +169,10 @@ def certificates_txt_dir(self) -> Path:
...
-        """
-        return self.auxiliary_datasets_dir / "maintenances"
-
-    @property
-    def mu_dataset_path(self) -> Path:
-        """
-        Returns a path to the dataset of maintenance updates
-        """
-        return self.mu_dataset_dir / "maintenance_updates.json"
-
```
J08nY, line 187: I need a way to get to these somehow, when handling the updates on the site. Can I do it with the new API? Iterate over aux_handlers and get their paths and detect based on type?

src/sec_certs/dataset/cc.py
```diff
@@ -256,12 +233,6 @@ def from_web_latest(cls) -> CCDataset:
     def _set_local_paths(self):
         super()._set_local_paths()
 
-        if self.auxiliary_datasets.pp_dset:
-            self.auxiliary_datasets.pp_dset.json_path = self.pp_dataset_path
-
-        if self.auxiliary_datasets.mu_dset:
-            self.auxiliary_datasets.mu_dset.root_dir = self.mu_dataset_dir
-
```
J08nY, line 264: Is this set properly in the new API?

src/sec_certs/dataset/cc.py
```diff
@@ -271,7 +274,13 @@ def _set_local_paths(self):
                 self.targets_txt_dir,
                 self.certificates_txt_dir,
             )
-        # TODO: This forgets to set local paths for other auxiliary datasets
+
+    def process_auxiliary_datasets(self, download_fresh: bool = False) -> None:
+        self.aux_handlers[CCMaintenanceUpdateDatasetHandler].certs_with_updates = [  # type: ignore
+            x for x in self if x.maintenance_updates
+        ]
+        self.aux_handlers[CCSchemeDatasetHandler].only_schemes = {x.scheme for x in self}  # type: ignore
```
J08nY, line 282: It seems a bit weird that these two handlers have special stuff here.

src/sec_certs/dataset/cc.py
```diff
@@ -780,164 +757,38 @@ def _compute_cert_labs(self) -> None:
...
+            if certified := scheme.lists.get(EntryType.Certified):
+                certs = [cert for cert in self if cert.status == "active"]
+                matches = CCSchemeMatcher.match_all(certified, scheme.country, certs)
+                for dgst, match in matches.items():
+                    self[dgst].heuristics.scheme_data = match
+            if archived := scheme.lists.get(EntryType.Archived):
+                certs = [cert for cert in self if cert.status == "archived"]
+                matches = CCSchemeMatcher.match_all(archived, scheme.country, certs)
+                for dgst, match in matches.items():
+                    self[dgst].heuristics.scheme_data = match
```
J08nY, line 832: Isn't this duplicate? I saw something like this function in the `heuristics` subpackage.

J08nY (commented):
Looks OK. But still has conflicts with main. Is the merge commit a real merge commit?

adamjanovsky:
> Looks OK. But still has conflicts with main. Is the merge commit a real merge commit?

Meh, something was left out, should be fixed by now.

src/sec_certs/dataset/cc.py
adamjanovsky: Fixed.

src/sec_certs/dataset/cc.py
adamjanovsky: Yes, see https://github.com/crocs-muni/sec-certs/blob/e6cab194fad1b924b060cf91f4b39e13b1cd55a1/src/sec_certs/dataset/dataset.py#L238-L240

src/sec_certs/dataset/cc.py
adamjanovsky: They form a dictionary and each child of `AuxiliaryDatasetHandler` must implement:
- `root_dir` attribute, where the dataset files are stored.
- `dset_path` attribute, where the final json is stored. 

Example, in the case of `CPEDataset` you can:

```python
cpe_dset = cc_dset.aux_handlers[CPEDatasetHandler].dset
cpe_dset_json_path = cc_dset.aux_handlers[CPEDatasetHandler].dset_path
```

src/sec_certs/dataset/cc.py
adamjanovsky: Not sure how to take care of this differently. The Handlers cannot access the `Dataset` instance they're bound to. And if they must be provided with some specific configuration (e.g., to avoid downloading data for all schemes, when only some are in the dataset), this information must be injected into the handlers.

adamjanovsky:
Hey, the initial draft of the functionality is implemented. Some notes below.

## Sample usage

### Create and fully process PP dataset

```python
pp_dset = ProtectionProfileDataset(root_dir="/path/to/pp/directory")
pp_dset.get_certs_from_web()
pp_dset.process_auxiliary_datasets()
pp_dset.download_all_artifacts()
pp_dset.convert_all_pdfs()
pp_dset.analyze_certificates()
```

### Acess to PP Dataset from CC Dataset

- A path to PP already-processed PP dataset can be provided in 

```python
cc_dset.process_auxiliary_dataset(processed_pp_dataset_root_dir="/path/to/pp/directory)
```

In such case, `PPDataset` instance will not be fully processed, but just copied into `cc_dset.auxiliary_datasets_dir`. The instance can be accessed via `cc_dset.aux_handlers[ProtectionProfileDatasetHandler].dset` once `cc_dset.process_auxiliary_datasets()` completes. 

Alternatively, `CCDataset` instance is capable of invoking full processing of `PPDataset` (it does so similarly with Maintenance updates) when `process_auxiliary_datasets()` is called without any argument.

When `CCDataset` processing is complete, the `ProtectionProfile` instances linked to specific `CCCertificate`s are listed as digests in `cert.heuristics.protection_profiles`. 

## Some notes on PP processing

- Some of collaborative PPs have multiple certification reports. I parsed only a single link.
- Some identical PPs are certified under multiple schemes. Ignoring, taking into account only a single scheme now
- CSV files not parsed. Only data entry missing is expected archival date of active PPs.
- PP ID not computed.
- Collaborative PPs with SDs pending review for compliance with the CC/CEM are not processed. 
- Maintenance updates of PP are left unprocessed

## Next steps

- @adamjanovsky: Write tests for PP Processing (soonish)
- @J08nY: Test the current design with web, propose changes or start integrating.

src/sec_certs/configuration.py
```diff
@@ -54,6 +54,10 @@ class Configuration(BaseSettings):
         "https://sec-certs.org/cc/cc.tar.gz",
         description="URL from where to fetch the latest full archive of fully processed CC dataset.",
     )
+    pp_latest_full_archive: AnyHttpUrl = Field(
+        "https://sec-certs.org/cc/pp.tar.gz",
+        description="URL from where to fetch the latest full archive of fully processed PP dataset.",
+    )
```
J08nY, line 60: The `pp_latest_snapshot` config also needs to change. It will no longer live on the `/static/` subdir. But have the same layout as the CC and FIPS datasets. Could you make the change pls?

src/sec_certs/constants.py
```diff
@@ -27,6 +27,44 @@
...
+    "Mobility",
+    "Multi-Function Devices",
+    "Network and Network-Related Devices and Systems",
+    "Operating Systems",
+    "Other Devices and Systems",
+    "Products for Digital Signatures",
+    "Trusted Computing",
+]
+
+CC_PORTAL_BASE_URL = "https://www.commoncriteriaportal.org"
```
J08nY, line 66: I remember seeing a nuch of URLs to CC portal stuff duplicated (I guess in the CC sample and dataset class). Is it now unified under subpaths of this base url? It is a nice time to unify.

src/sec_certs/dataset/auxiliary_dataset_handling.py
```diff
@@ -25,17 +24,25 @@
 
 
 class AuxiliaryDatasetHandler(ABC):
-    def __init__(self, root_dir: str | Path) -> None:
-        self.root_dir = Path(root_dir)
+    RELATIVE_DIR: ClassVar[str | None] = None
```
J08nY, line 27: Why the all caps?

src/sec_certs/dataset/cc.py
```diff
@@ -273,12 +272,34 @@ def _set_local_paths(self):
...
+                    f"Overwriting PP Dataset at {self.aux_handlers[ProtectionProfileDatasetHandler].root_dir} with dataset from {processed_pp_dataset_root_dir}."
+                )
+            shutil.copytree(
+                processed_pp_dataset_root_dir,
+                self.aux_handlers[ProtectionProfileDatasetHandler].root_dir,
+                dirs_exist_ok=True,
+            )
+
+        if skip_schemes:
+            del self.aux_handlers[CCSchemeDatasetHandler]
```
J08nY, line 301: Is it a good idea for it to delete the handler? It modifies the dataset state and I think that is quite unexpected.

src/sec_certs/dataset/cc.py
```diff
@@ -479,7 +493,7 @@ def _get_primary_key_str(row: Tag):
                 x.st_link,
                 None,
                 None,
-                profiles.get(x.dgst, None),
+                None,
```
J08nY, line 496: Why the change? Does this mean we will have no links to PPs in the df?

tests/data/cc/dataset/toy_dataset.json
```diff
@@ -226,7 +222,9 @@
           "indirectly_referenced_by": null,
           "indirectly_referencing": null
         },
-        "scheme_data": null
+        "scheme_data": null,
+        "protection_profiles": null,
+        "eal": null
```
J08nY, line 227: Why are these entries `null`, while the `"protection_profile_links"` entry has a single PP element? Is it that the toy dataset was not processed fully?

src/sec_certs/dataset/cc.py
J08nY: Makes sense.

requirements/dev_requirements.txt
```diff
@@ -2,6 +2,7 @@ accessible-pygments==0.0.4
     # via pydata-sphinx-theme
 aiohappyeyeballs==2.4.0
     # via aiohttp
+aiohttp==3.10.11
```
J08nY, line 5: This merge commit messed this up a bit. Same for the vulnerabilities notebook, but that is now fixed.

J08nY:
Could you please add some test(s) that run the PP pipeline at least once? I.e. improve the coverage in the
`dataset/protection_profile.py` and `sample/protection_profile.py` files [here](https://app.codecov.io/gh/crocs-muni/sec-certs/pull/466?dropdown=coverage&src=pr&el=h1&utm_campaign=pr+comments).

src/sec_certs/constants.py
adamjanovsky: Yes, I already did some unification, will double check!

src/sec_certs/dataset/auxiliary_dataset_handling.py
adamjanovsky: It's a constant

src/sec_certs/configuration.py
adamjanovsky: Sure, I wanted to discuss this first before changing this.

src/sec_certs/dataset/cc.py
adamjanovsky: Thanks for the tip. I needed something to commit on yesterday, I want to revisit this. It's a bad idea and I'm thinking about better design atm.

src/sec_certs/dataset/cc.py
adamjanovsky: Hm, I will enrich the DF with one extra attribute for that!

tests/data/cc/dataset/toy_dataset.json
adamjanovsky: Yes. Is in a state after `get_certs_from_web()` IMO. These are heuristics attributes filled-in only in the `analyze_dset()` step.

requirements/dev_requirements.txt
adamjanovsky: Yep, sorry for that. I'll bump the reqs once more before merging, treating also some open PRs.

adamjanovsky:
> Could you please add some test(s) that run the PP pipeline at least once? I.e. improve the coverage in the `dataset/protection_profile.py` and `sample/protection_profile.py` files [here](https://app.codecov.io/gh/crocs-muni/sec-certs/pull/466?dropdown=coverage&src=pr&el=h1&utm_campaign=pr+comments).

Sure 🙃 , see 

> @adamjanovsky: Write tests for PP Processing (soonish)

adamjanovsky:
## Regression tests

```markdown
OLD APPROACH
	- After `get_certs_from_web()`
--------------------`
	- # PP-rich certs: 3288
	- # PP links: 4291
	- # Unique PP links: 269
--------------------
	- # PP-rich certs: 3288
	- # PP links: 4291
	- # Unique PP links: 269
```

```markdown
NEW APPROACH
	- After `get_certs_from_web()`
--------------------
	- # PP-rich certs: 3259
	 -# PP links: 4292
	- # Unique PP links: 266
	- After processing ProtectionProfileDataset
--------------------
	- # PP-rich certs: 3212
	 -# PP links: 4232
	- # Unique PP links: 264
```

The decreased number of PPs in the new approach is explained by stronger requirements on linking from CCs to PPs. We require exact match of PP pdf link atm.

adamjanovsky:
This also implements #462, #463, #465

adamjanovsky:
@J08nY the requested changes should now be incorporated. I will further check that notebooks work. I will also flag the `NVDDatasetBuilder` tests with `xfail`, as they often fail due to NIST server problems.

J08nY:
> You can start working on the integration. I suggest merging this only once we're confident we can deploy this to sec-certs.org.

I guess we could do that, but my workflow kind of requires that the web is developed against the main branch here, but I will see. I mean real issues will only be visible after deploy anyway.

J08nY:
Could you make the pp URLs have /pp/ in the url? To align with the other blueprints on the web: /cc/ and /fips/.

adamjanovsky:
> > You can start working on the integration. I suggest merging this only once we're confident we can deploy this to sec-certs.org.
> 
> I guess we could do that, but my workflow kind of requires that the web is developed against the main branch here, but I will see. I mean real issues will only be visible after deploy anyway.

If we merge to main asynchronously, then we have broken API for a while. I would appreciate if you try to deploy from here. 

Once it all works, we can merge to main and create a new release.

J08nY:
What do you mean by broken API? What exactly breaks?

adamjanovsky:
> What do you mean by broken API? What exactly breaks?

Well, `ProtectionProfileDataset.from_web()` will not work till you deploy. Also, other classes will have incompatible data type till we run a new run on web.

adamjanovsky:
Hi, final batch of updates is here. 

- I joined two different methods `from_web_latest()` and `from_web()` in various classes. 
    - `from_web()` without additional arguments now defaults to `from_web_latest()`
- I added example pp notebook
- I added docs for pps
- I added support for pp processing in CLI

@J08nY I consider this done from my side. I'll be just assisting with integration and implementation of review requests.

src/sec_certs/sample/fips_iut.py
```diff
@@ -147,7 +147,10 @@ def from_web(cls) -> IUTSnapshot:
         """
         Get an IUT snapshot from the FIPS website right now.
         """
-        iut_resp = requests.get(constants.FIPS_IUT_URL)
+        if config.preferred_source_remote_datasets == "origin":
+            iut_resp = requests.get(constants.FIPS_IUT_URL)
+        else:
+            iut_resp = requests.get(config.fips_iut_dataset)
```
J08nY, line 153: I do not understand this change. We always had two methods:
 - `from_web`: This downloads/builds the dataset from the upstream/ground truth, so the CC portal, NIST page, etc.
 - `from_web_latest`: This downloads the processed dataset from our site.

Why did you decide to merge these two and use a configuration option to pick between them? Having functional, **runtime** control over which one gets used is quite important and I think this is a step in the wrong direction. The docstring is now also wrong (and assumptions about what this method does, elsewhere in to code or the website code are also all wrong now).

Please reconsider, I really do not like such a change inside a PP processing PR.

J08nY (changes requested):

src/sec_certs/sample/fips_iut.py
J08nY: I had a bit more of a look and I am okay with the change in the dataset API, but not in the sample API here.
(FIPS MIP and FIPS IUT). I believe these two should not be collapsed like this into one config-controlled method.

Also, the config and documentation is confused on what the name of the config value is:
`preferred_source_remote_datasets` vs `preferred_source_aux_datasets: "api" # set to "sec-certs" to fetch them from sec-certs.org` in the `user_guide.md` also, the values used in the config are "origin" vs "sec-certs" and not "api".

src/sec_certs/sample/document_state.py
```diff
@@ -0,0 +1,61 @@
+from dataclasses import dataclass
+from pathlib import Path
+
+from sec_certs.serialization.json import ComplexSerializableType
+
+
+@dataclass
+class DocumentState(ComplexSerializableType):
```
J08nY, line 8: DocumentState should likely be also used in the FIPS certificate for the security policy, right? Is now a good time to add it or no?

adamjanovsky: See https://github.com/crocs-muni/sec-certs/issues/464.

src/sec_certs/sample/fips_iut.py
adamjanovsky: Let's talk about this. IMO, having two public methods `from_web()` and `from_web_latest()` is somewhat confusing. I agree that MIP and IUT implementation has to change w.r.t. your requirements. 

My approach would be:

1. Create a [Protocol](https://typing.readthedocs.io/en/latest/spec/protocol.html) for `DownloadableDataset` that will force `from_web()` method. 
2. When called without any arguments, the `from_web()` method should choose URL based on configuration.
3. Explicit argument provided at runtime overrides such behaviour. 

Config doc should be changed.

src/sec_certs/sample/document_state.py
J08nY: Oh, I even saw that but forgot.

src/sec_certs/sample/fips_iut.py
J08nY: I think the current implementation for all classes apart from the FIPS IUT and FIPS MIP sample classes is fine. However, for those, there is a bit of a naming conflict that I remember from when I created them. The from_web is actually meant to download the sample from the **NIST web**, not our web. That is why it was called that, to align with other methods of that class, like `from_page` which creates the sample from the HTML if you have downloaded it some other way. So how about this (for the two classes):
 - They will have a `from_web` method that aligns with the rest of the framework in that it downloads the data from *our* web. It uses the config values pointing to the snapshot URLs by default on our web.
 - The method to download and parse the **NIST web** will be renamed, to something like `from_nist`.

Also, I just now noticed that the code as it is in the PR for the `from_web` method for the two classes does not make sense. The classes are IUT/MIP *sample* classes, yet it uses the config URL for IUT/MIP *dataset*. It should use the "*_latest_snapshot" config entry.

J08nY: I can commit this change and the fix for the config value if you want.

adamjanovsky: Oh gee, I didn't even notice I'm working with *samples*, not with *datasets*.

> So how about this (for the two classes):
> - They will have a from_web method that aligns with the rest of the framework in that it downloads the data from our web. It uses the config values pointing to the snapshot URLs by default on our web.
> - The method to download and parse the NIST web will be renamed, to something like from_nist.

makes perfect sense.

> I can commit this change and the fix for the config value if you want.

That'd be breat 👍 .

J08nY:
I am going to create a new PR based on a branch n which I rebased all of this on top of the current main to get rid of the messed up merge commit and have nicer history.

PR #473, opened 2025-02-13
Pre-release fixes

adamjanovsky:
- Fixes bad class name displayed in download progress bar
- Fixes typos in ProtectionProfile example notebook

### DIFF
--- a/notebooks/examples/protection_profiles.ipynb
+++ b/notebooks/examples/protection_profiles.ipynb
@@ -8,9 +8,9 @@
     "\n",
     "This notebook illustrated basic functionality of the `ProtectionProfileDataset` class that holds protection profiles bound to Common Criteria certified products. The object that holds a single profile is called `ProtectionProfile`. \n",
     "\n",
-    "Note that there exists a front end to this functionality at [sec-certs.org/cc](https://sec-certs.org/cc/). Before reinventing the wheel, it's good idea to check our web. Maybe you don't even need to run the code, but just use our web instead. \n",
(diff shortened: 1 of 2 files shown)


adamjanovsky:
@J08nY I also tested `ProtectionProfileDataset.from_web(auxiliary_datasets=True, artifacts=True)`. It fails, because the output folder is populated with `pp.json` file and the `from_web()` method then attempts to load the dataset object from `dataset.json`, resulting in `FileNotFoundError`. I guess this should be fixed on your end. Fixing this in lib would imply handling this individually for different datasets.

In parallel, I've tested `FIPSDataset` download. It works.

J08nY:
Hmm, I think this is because the auxiliary dataset handler for PPs uses the pp.json name, I think I got it from there. I will check, but in that case the unification should be done there.

J08nY:
Yeah so the core of the issue was here: https://github.com/crocs-muni/sec-certs/blob/bdbd974847b713d14ccd90b617bae4e41447c09e/src/sec_certs/dataset/auxiliary_dataset_handling.py#L259-L261

The PP dataset expects a different name when it is an auxilliary dataset vs when it is standalone.

adamjanovsky:
Ok, changed that to `dataset.json`. Note that the `from_web()` will still not work till updated on web.

J08nY:
> Ok, changed that to `dataset.json`. Note that the `from_web()` will still not work till updated on web.

Ok, the next update will fix that. I think we can release and I will update on the web.

PR #479, opened 2025-02-15
Get rid of DUMMY_NONEXISTING_PATH

### LINKED ISSUE #290: None instead of DUMMY_NONEXISTING_PATH when path not set on Dataset instances
Currently, objects that inherit from `Dataset` instance can be initialized without `root_dir` properly set. In such case, its value is set to `DUMMY_NONEXISTING_PATH` (constants module) which is a path object meant to substitute `None` value. 

When any operation that involves serialization occurs on the dataset, it is checked that `root_dir` was properly set by `@serialize` decorator which raises an exception in case it was not yet set. 

What we're trying to achieve is basically `None` behaviour that could be somewhat tricky to handle with MyPy. ATM, we take the advantage of `root_dir` always being a `Path` instance. We should, however, replace `DUMMY_NONEXISTING_PATH` with `None` in the future. For some rationale, see https://github.com/crocs-muni/sec-certs/pull/275/commits#discussion_r1042430764

J08nY:
Fixes #290.

### DIFF
--- a/src/sec_certs/dataset/auxiliary_dataset_handling.py
+++ b/src/sec_certs/dataset/auxiliary_dataset_handling.py
@@ -178,7 +178,7 @@ class FIPSAlgorithmDatasetHandler(AuxiliaryDatasetHandler):
 class CCSchemeDatasetHandler(AuxiliaryDatasetHandler):
     def __init__(
         self,
-        aux_datasets_dir: str | Path = constants.DUMMY_NONEXISTING_PATH,
+        aux_datasets_dir: str | Path | None,
         only_schemes: set[str] | None = None,
     ):
         super().__init__(aux_datasets_dir)
@@ -209,7 +209,7 @@ class CCMaintenanceUpdateDatasetHandler(AuxiliaryDatasetHandler):
 
     def __init__(
         self,
-        aux_datasets_dir: str | Path = constants.DUMMY_NONEXISTING_PATH,
+        aux_datasets_dir: str | Path | None,
         certs_with_updates: Iterable[CCCertificate] | None = None,
     ) -> None:
         super().__init__(aux_datasets_dir)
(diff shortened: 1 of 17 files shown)


J08nY:
@adamjanovsky What do you think about this? It gets rid of the `DUMMY_NONEXISTING_PATH`. The cost is a few ignores in typing. However, I would argue that those are quite reasonable, as before the distinction of whether something is a dummy path or not was not done on type level anyway. So on a static level we do not gain anything and on a runtime level it is a bit more clear what is going on and we have no chance to get errors where we try to write somewhere we shouldn't.

src/sec_certs/dataset/dataset.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         self.name = name if name else type(self).__name__
         self.description = description if description else datetime.now().strftime("%d/%m/%Y %H:%M:%S")
         self.state = state if state else self.DatasetInternalState()
-        self.root_dir = Path(root_dir)
+        self.root_dir = Path(root_dir) if root_dir is not None else None  # type: ignore
```
adamjanovsky, line 68: Why is the `# type: ignore` required? Shouldn't `self.root_dir` expect `Path | None` assignments?

src/sec_certs/dataset/auxiliary_dataset_handling.py
```diff
@@ -30,8 +29,8 @@ class AuxiliaryDatasetHandler(ABC):
     aux_datasets_dir: Path
```
adamjanovsky, line 29: Just set this to `Path |  None` to avoid the type warnings?

adamjanovsky (commented):
Sure, that's a step in a good direction. I provided some individual comments. I also think that a release would be nice once you're finished with these tiny-refactoring PRs.

src/sec_certs/serialization/json.py
```diff
@@ -113,6 +126,23 @@ def _serialize(*args, **kwargs):
...
 
+def only_backed(throw: bool = True):
+    def deco(func: TCallable) -> TCallable:
+        @wraps(func)
+        def _only_backed(*args, **kwargs):
+            if args[0].root_dir is None:
+                if throw:
+                    raise ValueError(f"Method {func.__name__} can only be called on backed dataset.")
+                else:
+                    return None
```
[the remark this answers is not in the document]
J08nY: You mean in the `None` branch? Sure will add that.

src/sec_certs/dataset/auxiliary_dataset_handling.py
J08nY: Well, we need to lie here as well I think. The same way we are lying in the `Dataset.root_dir` property claiming that it is never `None` and only protecting it via the `only_backed` decorator. Otherwise, typewise it just all goes wrong and every access to any attribute derived from `root_dir` needs to be checked and guarded.

Issue #480, opened 2025-02-25
Annotating certificate references with LLMs


fedorst:
We should have a proof-of-concept of annotating certificate references (component_used, previous_version, etc) with LLMs.

J08nY:
This is actually closed due to #481.

PR #481, opened 2025-02-25
Demonstration notebook of annotation of references with LLMs

fedorst:
#480


adamjanovsky:
Hi, thanks for this effort 👍. Few tips:

- ATM, we throw the whole report text at the LLM. Could we attempt to limit the content (based on which the prediction is made) to some surrounding sentences of the referenced certificate ID? Could this help? We have the corresponding functionality ready, provided by

https://github.com/crocs-muni/sec-certs/blob/27abe1d969703a87aace43334969939be1b8c9f8/src/sec_certs/model/references_nlp/segment_extractor.py#L155

- IMO it makes sense to compare the performance on the metrics used in the paper (so F1 score and ROC-AUC). What do you think?
- Ideally, we'd like the notebook to be moved to `notebooks/cc` folder, as it relates to Common Criteria. 
- After extending `pyproject.toml` dependencies, could you please `cd requirements && ./compile.sh` to create a pinned list of updated dependencies that you will subsequently commit? (There are some additional contributing tips, see https://sec-certs.org/docs/contributing.html#requirements)
- Other than that, this is a very nice contribution. It shows us how to use OpenAI API with our artifacts and confirms that LLM does not improve the reference context inference out of the box.

adamjanovsky (commented):
Hi, thanks for this effort 👍. Few tips:

- ATM, we throw the whole report text at the LLM. Could we attempt to limit the content (based on which the prediction is made) to some surrounding sentences of the referenced certificate ID? Could this help? We have the corresponding functionality ready, provided by

https://github.com/crocs-muni/sec-certs/blob/27abe1d969703a87aace43334969939be1b8c9f8/src/sec_certs/model/references_nlp/segment_extractor.py#L155

- IMO it makes sense to compare the performance on the metrics used in the paper (so F1 score and ROC-AUC). What do you think?
- Ideally, we'd like the notebook to be moved to `notebooks/cc` folder, as it relates to Common Criteria. 
- After extending `pyproject.toml` dependencies, could you please `cd requirements && ./compile.sh` to create a pinned list of updated dependencies that you will subsequently commit? (There are some additional contributing tips, see https://sec-certs.org/docs/contributing.html#requirements)
- Other than that, this is a very nice contribution. It shows us how to use OpenAI API with our artifacts and confirms that LLM does not improve the reference context inference out of the box.

adamjanovsky:
Also, @J08nY we still have some scheme tests failing, see log https://github.com/crocs-muni/sec-certs/actions/runs/13521639405/job/37782051287?pr=481. Could we adjust them to xfail?

fedorst:
Hey, 

>     * ATM, we throw the whole report text at the LLM. Could we attempt to limit the content (based on which the prediction is made) to some surrounding sentences of the referenced certificate ID? Could this help? We have the corresponding functionality ready, provided by

In principle we could probably save some money and time by doing that, however, sometimes we care not only about the referenced certificate ID but also the name of the entity being certified, or references to the name of the target certificate, which might not always be in the same place.

>     * IMO it makes sense to compare the performance on the metrics used in the paper (so F1 score and ROC-AUC). What do you think?

F1 - done. Can't give ROC-AUC because I don't have thresholdable numeric values. Also note that this all is on the training set (train.csv) - but I am not actually using it for training anything - this is all just in-context learning based on the examples in the codebook and written instructions.

>     * Ideally, we'd like the notebook to be moved to `notebooks/cc` folder, as it relates to Common Criteria.

Done.

>     * After extending `pyproject.toml` dependencies, could you please `cd requirements && ./compile.sh` to create a pinned list of updated dependencies that you will subsequently commit? (There are some additional contributing tips, see https://sec-certs.org/docs/contributing.html#requirements)

Done.

> confirms that LLM does not improve the reference context inference out of the box.

Yes, but also when reading the examples, the categorizations did not make sense to me a lot of the time - so I think there is a lot of noise in the data by default. A clearly written-out hierarchy of when to prefer what label given overlap would have been useful. The codebook could have included a brief justification of "why" a particular secondary label was chosen for an example text.

J08nY:
> > Hmm, they are all xfail, but apparently only for RequestException. Hmm I want to strike a balance between making the worlflows green on unrelated pullrequests and getting notified when the scheme stuff breaks so it can be fixed...
> 
> The tests can break because:
> 
> * a change is made on the scheme side (external to us)
> * a change is made in the code
> 
> The first can be captured by scheduled pipelines that would run these tests selectively. The second cause could be addressed by monitoring the files subject to a specific commit and triggering the test only if these concern scheme code. This has caveats though, as e.g., modified utility funcions would not trigger the tests. Also, it adds some complexity.

This applies to all tests that use remote resources, so even some CC and FIPS/NVD tests. What about the following:
 - Add a custom mark to these tests (e.g. `remote`).
 - Do not run the tests by default (similar to "slow" tests) for PRs and the like.
 - Run a weekly test of the remote tests, using [scheduled](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) workflows.
Wdyt?

Also, sorry for the offtopic discussion about this. Will create an issue for this.

J08nY:
> In principle we could probably save some money and time by doing that, however, sometimes we care not only about the referenced certificate ID but also the name of the entity being certified, or references to the name of the target certificate, which might not always be in the same place.

We have quite a bit more structured data already available about both certificates. We could present it somehow in the context. For example: If the referencing certificate has a category of "printers" and the referenced certificate has a category of "smartcard" it is unlikely that they have "previous_version" relationship.

> Yes, but also when reading the examples, the categorizations did not make sense to me a lot of the time - so I think there is a lot of noise in the data by default. A clearly written-out hierarchy of when to prefer what label given overlap would have been useful. The codebook could have included a brief justification of "why" a particular secondary label was chosen for an example text.

We could provide such reasoning in text (we discussed some of it offline already) if it would help in the classification. However, I wonder, if we kept it a bit more open-ended and asked about the LLM about the relationship of the two certified products what would we get?

adamjanovsky:
> Hey,
> 
> > ```
> > * ATM, we throw the whole report text at the LLM. Could we attempt to limit the content (based on which the prediction is made) to some surrounding sentences of the referenced certificate ID? Could this help? We have the corresponding functionality ready, provided by
> > ```
> 
> In principle we could probably save some money and time by doing that, however, sometimes we care not only about the referenced certificate ID but also the name of the entity being certified, or references to the name of the target certificate, which might not always be in the same place.
> 
> > ```
> > * IMO it makes sense to compare the performance on the metrics used in the paper (so F1 score and ROC-AUC). What do you think?
> > ```
> 
> F1 - done. Can't give ROC-AUC because I don't have thresholdable numeric values. Also note that this all is on the training set (train.csv) - but I am not actually using it for training anything - this is all just in-context learning based on the examples in the codebook and written instructions.
> 
> > ```
> > * Ideally, we'd like the notebook to be moved to `notebooks/cc` folder, as it relates to Common Criteria.
> > ```
> 
> Done.
> 
> > ```
> > * After extending `pyproject.toml` dependencies, could you please `cd requirements && ./compile.sh` to create a pinned list of updated dependencies that you will subsequently commit? (There are some additional contributing tips, see https://sec-certs.org/docs/contributing.html#requirements)
> > ```
> 
> Done.
> 
> > confirms that LLM does not improve the reference context inference out of the box.
> 
> Yes, but also when reading the examples, the categorizations did not make sense to me a lot of the time - so I think there is a lot of noise in the data by default. A clearly written-out hierarchy of when to prefer what label given overlap would have been useful. The codebook could have included a brief justification of "why" a particular secondary label was chosen for an example text.

Thanks, I think we clarified all the stuff in person today.

fedorst:
Running it on segments yields a slightly better result, however, the main diagonal is still zeroes except for component_used.

adamjanovsky:
> Running the classification on segments with subsequent voting yields a slightly better result, however, the main diagonal is still zeroes except for component_used.

Understood. I suppose we merge with this as a conclusion?

PR #491, opened 2025-03-05
Semantic search demo

fedorst:
# Certificate Vector Search Demo

This is a single-file Python script that demonstrates how to process and search through certificate documents using vector embeddings. It’s not a full-blown project but rather a proof-of-concept for semantic search over certificate datasets. It’s  simple and hopefully readable, but not super scalable out of the box.

Assumes having the target folder, in my case `cc_data_gemini` in the root directory of the project. Assumes in turn that it contains `dataset.json` and a folder structure with `.txt` files. Note that embedding will take a while if you will do it on the entire dataset. I ran it like this:
```sh
python vector_search.py --data-path ../../cc_data_gemini --db-path vector_db.sqlite --force-rebuild --port 8080
```

Screenshot:
image

---

See rest of the details [here](https://github.com/crocs-muni/sec-certs/blob/8b10aede63e68ed9c0a7f37be9f6b4dc8044fc3e/notebooks/cc/vector_search.md)

### DIFF
--- a/notebooks/cc/vector_search.md
+++ b/notebooks/cc/vector_search.md
@@ -1,0 +1,106 @@
+# Certificate Vector Search Demo
+
+This is a single-file Python script that demonstrates how to process and search through certificate documents using vector embeddings. It’s not a full-blown project but rather a proof-of-concept for semantic search over certificate datasets. It’s designed to be simple and accurate, though not highly scalable out of the box.
+
+---
+
+## What It Does
+
+1. **Processes Certificate Documents**:
+   - Takes a dataset of certificates (loaded from a JSON file).
+   - Extracts text from associated `.txt` files (e.g., reports, security targets, certificates).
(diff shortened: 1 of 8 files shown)


adamjanovsky:
Hi, that's great, thanks for this effort 👍 . 

- Do you happen to know how long does it take to embedd the documents on your machine? (Something more precise than "a while" would be great :D)
- Could you please fix the Lint pipeline errors? Tutorial on running these tools locally is at https://sec-certs.org/docs/contributing.html#quality-assurance
- During our final meeting we should discuss what are the trade-offs of using sentence transformers vs. token-based transformers. We have some queries and some evaluation dataset, so we could quite easily annotate the quality and compare it with token-based transformers as utilized in [D. Macko's Master's thesis](https://is.muni.cz/auth/th/k87j7/MasterThesis.pdf)

I think we can merge subsequently.

adamjanovsky (approved):
Fedor, let me thank you once more for your assistance during fellowship 👍 . I just merged fresh main and resolve some dependency conflicts. Approving and merging. Have a nice time!

PR #493, opened 2025-04-02
Improve BSI cert_id regex.

J08nY:


petrs:
Thank you Jano - is this WIP or shall we merge it already?

PR #505, opened 2025-06-12
Add version warning for dataset compatibility

### LINKED ISSUE #338: Start collecting dataset version
When collecting a CC/FIPS dataset, we could collect the tool versions used to produce such datasets. When the user fetches `from_web_latest()`, we could display a warning when different version is run. We could assist the user by displaying him how to install the tool of that specific release. E.g., `pip install sec-certs xxx`. 

Care must be taken when dealing with datasets created with commit x, but updated on commit y.

## TODO

- Create a new attribute `sec_certs_version` of `DatasetInternalState` and populate it with version obtained from `setuptools_scm` when a dataset is created. 
- Design it so that the attribute is filled only when dataset is created, not when dataset is updated, de/serialized, etc.
- When a dataset is loaded, compare the current tool version with the dataset version. If mismatch, log warning.

J08nY:
Introduce a version check for datasets to warn users if the dataset version differs from the current package version. Update tests to accommodate the new version handling.

Fixes #338.

Generated with Github Copilot agent.

### DIFF
--- a/src/sec_certs/dataset/dataset.py
+++ b/src/sec_certs/dataset/dataset.py
@@ -277,7 +277,7 @@ class Dataset(Generic[CertSubType], ComplexSerializableType, ABC):
             from sec_certs._version import __version__ as current_version
         except ImportError:
             current_version = "unknown"
-        dset_version = getattr(getattr(dset, 'state', None), 'sec_certs_version', None)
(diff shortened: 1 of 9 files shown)


adamjanovsky:
Hehe, makes sense.

PR #513, opened 2025-08-24
feat: Prototype with dropdown menu for dynamically adding graph with callback

xvalec01:
This PR brings an update to the feature branch with part of the functionality of the dashboarding system.
Until the feature branch is synced with the "page" branch, I will add the files of interest.
```
┣ dashboard/
┃ ┃ ┣ docs/
┃ ┃ ┃ ┗ prettier_class_diagram.puml
┃ ┃ ┣ graphs/
┃ ┃ ┃ ┣ base.py
┃ ┃ ┃ ┣ cc_bar_chart.py
┃ ┃ ┃ ┣ cc_pie_chart.py
┃ ┃ ┃ ┗ registry.py
┃ ┃ ┣ callback.py
┃ ┃ ┣ cc_route.py
┃ ┃ ┣ data.py
┃ ┃ ┣ fips_route.py
┃ ┃ ┣ layout.py
┃ ┃ ┗ __init__.py
```


sec_certs_page/dashboard/__init__.py
```diff
@@ -1,20 +1,33 @@
-# dashboard
+"""Initializes the Dash dashboard application and its components."""
+
 from flask import Flask
 from flask_wtf import CSRFProtect
 
+from sec_certs_page import mongo
```
J08nY, line 6: Relative import please.

sec_certs_page/dashboard/__init__.py
```diff
@@ -1,20 +1,33 @@
...
+
 from ..common.dash.base import Dash
```
J08nY, line 8: This could be moved to /dashboard, no  need for it to remain in common if it will not be used by other bluprints.

sec_certs_page/dashboard/graphs/cc_bar_chart.py
```diff
@@ -0,0 +1,60 @@
+"""A concrete implementation of a bar chart for CC certificates by manufacturer."""
+
+import plotly.graph_objects as go
+from dash import Dash, dcc, html
+from dash.dependencies import Input, Output
+
+from .base import BaseGraph
+
+
+class CCBarChart(BaseGraph):
```
J08nY, line 10: Also, this is not a good name. The fact that it is a bar chart is secondary. The fact that it is certificates by category over year is important.

sec_certs_page/dashboard/graphs/registry.py
```diff
@@ -0,0 +1,44 @@
+"""Implements the Registry Pattern for managing dashboard graphs."""
+
+from typing import Dict, Iterator
+
+from sec_certs_page.dashboard.graphs.base import BaseGraph
```
J08nY, line 5: Relative import please.

sec_certs_page/dashboard/data.py
```diff
@@ -0,0 +1,56 @@
...
+                data = list(cursor)
+                if data:
+                    self._df_fips = pd.DataFrame(data)
+                else:
+                    logger.warning("FIPS collection is empty.")
+                    return pd.DataFrame()
+            except Exception as e:
+                logger.exception("Error fetching FIPS data from MongoDB")
+                raise e
+        return self._df_fips
```
J08nY, line 56: This will cache the dataframes for too long. Even if the underlying data changes in the meantime. Also, making a dataframe this way from the whole dataset is not efficient (idk what it even does to the very deep and branchig json structure).

sec_certs_page/dashboard/graphs/cc_bar_chart.py
xvalec01: Good point, will change it.

sec_certs_page/dashboard/graphs/registry.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class GraphRegistry:
```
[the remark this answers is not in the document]
J08nY: Ok makes sense. Let's see where it goes.

J08nY:
I think you can rebase this on top of current `page` branch, as the `feat/dash-page` got merged. I didn't really think before merging it, but now it is too late.

J08nY:
One general comment, which may seem nitpicky, but I would appreciate if you used "plot" instead of "graph" to describe the plots. The thing is that graph can be understood as a network/graph and not a plot. Within our context there is actually a reference graph/network so it may be misleading. So either plot or chart.

xvalec01:
> One general comment, which may seem nitpicky, but I would appreciate if you used "plot" instead of "graph" to describe the plots. The thing is that graph can be understood as a network/graph and not a plot. Within our context there is actually a reference graph/network so it may be misleading. So either plot or chart.

Okay, I will get rid of the "graph" across the dashboard part.

J08nY:
I have rebased this branch on top of page and pushed that into the [crocs-muni:feat/page-dash-update](https://github.com/crocs-muni/sec-certs/tree/feat/page-dash-update) branch.

J08nY:
Hey @xvalec01 what is the status on this?

PR #527, opened 2025-11-09
Add legal disclaimers

### LINKED ISSUE #526: Add clear, site-wide disclaimer for automated heuristics (esp. CVE-to-certificate matches) and explicit contact point
During the iSysel legal seminar, it was raised that manufacturers may face *reputational harm* due to how sec-certs.org currently presents heuristic outputs, particularly CVE matches to certified products. While we do mention the possibility of false positives/negatives near specific items, we’re *not sufficiently transparent* at a global level about the limitations of our automated methods and the uncertainty inherent in these outputs.

We should:
- Add a generic disclaimer covering all heuristic attributes (e.g., CVE matches, certificate clustering, etc.).
- Make explicit that these results are *automatically generated*, *not human-reviewed by default*, and should be treated as such.
- Refer readers to our methodology for details about data sources and heuristics.
- Provide a clear and explicit point of contact for dispute resolution or corrections.

## Problem Statement

Current per-item phrasing (examples):
 > “This certificate has known related CVEs, which means that the certified product might be vulnerable.”
 
> “The CVE was matched to the following certificates. This means the vulnerability likely affects said certified devices. However, this is an automated process that may have false positives and negatives.”

Issues:
- Disclaimers are localized and inconsistent.
- Language may overstate confidence (“likely affects”) while the method is heuristic.
- There is no single, prominent disclaimer covering the broader set of automated inferences.
- Contact/appeal path is not explicit.

## Proposed disclaimers

1. To be shown in Heuristics section of every cert.

> *Automated inferences — use with caution*  
> All attributes shown in this section (e.g., links between certificates, products, vendors, and known CVEs) are generated by *automated heuristics* and *have not been systematically reviewed by humans*. These methods can produce false positives and false negatives and should not be treated as definitive without independent verification.  
> For details on our data sources and inference methods, see our [Methodology](TBA). If you believe any information here is inaccurate or harmful, please [contact us](TBA) or [submit feedback](TBA). 

2. To be shown on every CVE page

CVE associations below are derived from automated matching. They may include errors. See the pages of the corresponding certificates and [Methodology](TBA) for details.

J08nY:
This tries to go towards #526.

Some examples of how this looks:

## Certificate page

image

image

## CVE page

image

## CPE page

image

Fix #526

### DIFF
--- a/sec_certs_page/templates/vuln/cpe.html.jinja2
+++ b/sec_certs_page/templates/vuln/cpe.html.jinja2
@@ -34,12 +34,7 @@
 
                 <div class="row pt-5">
                     <div class="col-12 col-md-6">
-                        <h2 class="anchor" id="data">Data
-                            <span style="font-size: 0.5em; vertical-align: middle" tabindex="0"
-                                  class="badge bg-secondary"
-                                  data-bs-toggle="tooltip" data-bs-placement="top"
-                                  title="Information extracted from CVE data from MITRE and NIST's NVD.">?</span>
-                        </h2>
+                        <h2 class="anchor" id="data">Data</h2>
                         <table class="table table-sm table-borderless w-auto">
                             <colgroup>
                                 <col style="width: 30%"/>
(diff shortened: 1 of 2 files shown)


adamjanovsky:
Looks nice. What do we fill the methodology with? I was thinking we could just link to our research papers...

PR #529, opened 2025-11-12
Ignore CVSS V4.0 metrics in CVE loading.

J08nY:
THe NVD no longer offers the "exploitabilityScore" and "impactScore" for CVSS V4.0 in their API:

https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema (See the "cvss-v40" entry)

We should figure out what to do as more CVEs will likely have this in the future and we will lack data. The cvssData scheme is here:

https://csrc.nist.gov/schema/nvd/api/2.0/external/cvss-v4.0.json

### DIFF
--- a/src/sec_certs/sample/cve.py
+++ b/src/sec_certs/sample/cve.py
@@ -27,19 +27,22 @@ class CVE(PandasSerializableType, ComplexSerializableType):
             """
             Loads metrics from dictionary
             """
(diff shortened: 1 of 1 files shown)


J08nY:
@adamjanovsky This broke CVE processing on the server like a week ago. We need to decide what to do in the future about the metrics.

PR #530, opened 2025-11-17
Move to uv

J08nY:
This started as a PR to bump the spacy requirement:

> This PR only limits spacy < 3.8.0 for aarch64 builds. This should avoid the issue in #475 while allowing more recent spacy.
> @adamjanovsky Could we investigate whether the issue described there still exists? There is for example https://github.com/explosion/spaCy/issues/13751 which says these issues should not happen anymore?

However, in doing so I hit a bunch of roadblocks with the `pip-compile` approach we are using. As discussed in dc8cd4b, pip-compile is not cross-platform or cross-python compatible. I ended up moving most things to `uv`.

The project: 
 - Still installs via `pip install sec-certs` or `pip install -e .`. This should be always supported and we should strive to keep the `pyproject.toml` dependencies specified such that this works on most platforms by default.
 - Locked dependencies are in `uv.lock` which is cross-platform and universal. I tested 3.10-3.12 for now. More should *just work*.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -18,7 +18,7 @@ jobs:
         run: sudo apt-get install -y build-essential libpoppler-cpp-dev pkg-config python3-dev
       - uses: actions/checkout@v4
       - name: Install uv and Python
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@v7
         with:
           python-version: ${{ matrix.python-version }}
           enable-cache: true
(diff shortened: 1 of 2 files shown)


J08nY:
I wanted to change the minimal amount of stuff and did not want to touch the release process if it works. But you are right that this may be a good moment to actually do that. I will have a look at it.

adamjanovsky:
> There is one more change here that I am unsure about. I merged the `test` and `dev` extras as they were kind of redundant. That I think is OK. However, then I saw that there is a feature called "[dependency-groups](https://packaging.python.org/en/latest/specifications/dependency-groups/)" which are meant for use cases like dev dependencies. Then I moved the new `dev` extra into a `dev` dependency group. However, that means that now: `pip install -e .[dev]` does not work anymore (and it did before). The syntax is: `pip install -e . --group dev` or `uv sync --dev`. What do you think?

That's ok, I don't think we need extra compatibility. Also, unless I'm mistaken, the `dev` group is installed by default, so just `uv sync` would work.

PR #531, opened 2025-11-18
feat: Page with dashboarding base functionality

xvalec01:
Okay, so this PR has:

Creating a dashboard that can have multiple charts. Each chart is defined by the filters that correspond to the chosen fields from the database. 

When you save a chart, it will be stored just in the current browser session. To save it for good, you must save the dashboard that this chart is part of. 

It is possible to delete the charts from the dashboard.

Everything is behind the authentication for users with the `admin` or `dashboard` role.

For now, the dashboard system is approachable with this dropdown in the navigation for such roles:
image
Later, I would use `Analysis` (or name it differently) in the dropdowns of the main navigation.

The visual design is far from perfect. Will upgrade it in an additional PR.

### DIFF
--- a/sec_certs_page/dashboard/data.py
+++ b/sec_certs_page/dashboard/data.py
@@ -336,16 +336,16 @@ class DataService:
 
         return df
 
-    def get_dataset_metadata(self, collection_name: str) -> tuple[dict[str, ColumnStats], int, int]:
+    def get_dataset_metadata(self, collection_name: CollectionName) -> tuple[dict[str, ColumnStats], int, int]:
         """Get metadata about a dataset including column information.
 
         :param collection_name: Type of dataset ('cc' or 'fips')
         :return: Tuple of (column_stats, total_records, total_columns) where
                  column_stats maps column names to their statistics
         """
-        if collection_name.lower() == "cc":
+        if collection_name.value.lower() == "cc":
             df = self.get_cc_dataframe()
-        elif collection_name.lower() == "fips":
+        elif collection_name.value.lower() == "fips":
             df = self.get_fips_dataframe()
         else:
             raise ValueError(f"Unknown dataset type: {collection_name}")
--- a/sec_certs_page/dashboard/pages/common.py
+++ b/sec_certs_page/dashboard/pages/common.py
@@ -100,7 +100,7 @@ def create_page_stores(collection_name: CollectionName) -> list:
     ]
 
 
-def create_dashboard_management_buttons(collection_name: str) -> dbc.Row:
+def create_dashboard_management_buttons(collection_name: CollectionName) -> dbc.Row:
     """
     Create the dashboard management buttons (Create New, Load Predefined).
 
@@ -143,7 +143,7 @@ def create_dashboard_management_buttons(collection_name: str) -> dbc.Row:
     )
 
 
-def create_empty_state(collection_name: str) -> html.Div:
+def create_empty_state(collection_name: CollectionName) -> html.Div:
     """
     Create the empty state display when no dashboard is selected.
 
@@ -175,7 +175,7 @@ def create_empty_state(collection_name: str) -> html.Div:
     )
 
 
-def create_chart_creation_modal(collection_name: str) -> dbc.Modal:
+def create_chart_creation_modal(collection_name: CollectionName) -> dbc.Modal:
     """
     Create the modal for custom chart creation.
 
@@ -482,7 +482,7 @@ def create_chart_creation_modal(collection_name: str) -> dbc.Modal:
     )
 
 
-def create_chart_controls(collection_name: str) -> dbc.Card:
+def create_chart_controls(collection_name: CollectionName) -> dbc.Card:
     """
     Create the chart selection and addition controls.
 
@@ -569,7 +569,7 @@ def create_chart_controls(collection_name: str) -> dbc.Card:
     )
 
 
-def create_dashboard_action_buttons(collection_name: str) -> dbc.Row:
+def create_dashboard_action_buttons(collection_name: CollectionName) -> dbc.Row:
     """
     Create the dashboard action buttons (Update All, Save).
 
@@ -609,7 +609,7 @@ def create_dashboard_action_buttons(collection_name: str) -> dbc.Row:
     )
 
 
-def create_dashboard_content(collection_name: str) -> html.Div:
+def create_dashboard_content(collection_name: CollectionName) -> html.Div:
     """
     Create the dashboard content area shown when a dashboard is active.
 


pyproject.toml
```diff
@@ -76,7 +74,8 @@
     "requests",
     "feedgen",
     "pendulum",
-  ]
+    "numpy<2",
```
J08nY: Lets drop this constraint, it should be unnecessary after the upstream dependency update and the move to uv.

sec_certs_page/common/dash/base.py
```diff
@@ -17,6 +17,7 @@ def interpolate_index(
         app_entry="",
         favicon="",
         renderer="",
+        **kwargs,
```
J08nY: This file can be moved to the dashboard subpackage, no need to keep it here.

sec_certs_page/dashboard/callbacks/charts.py
```diff
@@ -0,0 +1,232 @@
...
+) -> None:
+    """Must be registered once globally since pattern-matching callbacks match across all collections."""
+
+    @dash_app.callback(
+        output=dict(content=Output({"type": "chart-content", "index": MATCH}, "children")),
+        inputs=dict(n_clicks=Input({"type": "chart-refresh", "index": MATCH}, "n_clicks")),
+        state=dict(wrapper_id=State({"type": "chart-wrapper", "index": MATCH}, "id")),
+        prevent_initial_call=True,
+    )
+    def refresh_single_chart(n_clicks, wrapper_id):
```
J08nY: Maybe a docstring here?
```suggestion
    def refresh_single_chart(n_clicks, wrapper_id):
        """When user clicks the chart-refresh button we re-render it."""
```

sec_certs_page/dashboard/callbacks/dashboard.py
```diff
@@ -0,0 +1,332 @@
...
+
+def _register_dashboard_names(
+    dash_app: "Dash",
+    prefix: str,
+    dataset_type: CollectionName,
+    dashboard_manager: "DashboardManager",
+) -> None:
+    @dash_app.callback(
+        output=dict(options=Output(f"{prefix}-dashboard-selector", "options")),
+        inputs=dict(loaded=Input(f"{prefix}-dashboard-loaded", "data")),
```
J08nY: This is a general comment about these dash callbacks and inputs and outputs. I find it quite hard to see where they are all connected to. The best I found was to search the codebase for the name and then try to understand the context.

I am thinking if we could improve this somehow. Most of them are prefixed by the collection name to distinguish them. So they are kind of "scoped" to the collection, but really they are scoped to the "Dash page" right? It is just that currently page == collection. What about having like a class like:

```python
class Collection
    def __init__(self, prefix: str):
        self.prefix = prefix

    @property
    def dashboard_selector(self):
        return f"{self.prefix}-dashboard-selector"
```

Then in the various inputs/outputs you could do `Collection(prefix).dashboard_selector` and this would be immediately navigable through the codebase via IDEs and such. The class would just have a bunch of properties. Optionally, it could also return not just the string but some object which would give you the `Output(string, type)` or `Input(string, type)` so you would do:

```python
Collection(prefix).dashboard_selector.output
```

and it would actually resolve to `Output(f"{prefix}-dashboard-selector", "options")`.

So to summarize, this comment raises two questions:
 - How are these callbacks inputs "scoped" actually, why they need to differ between CC and FIPS?
 - Can we make them easier to lookup in the codebase by wrapping these string-id creations somehow.

sec_certs_page/dashboard/callbacks/dashboard.py
```diff
@@ -0,0 +1,332 @@
...
+    @dash_app.callback(
+        output=dict(
+            selector_value=Output(f"{prefix}-dashboard-selector", "value"),
+            loaded=Output(f"{prefix}-dashboard-loaded", "data"),
+        ),
+        inputs=dict(collection_name=Input(f"{prefix}-collection-name", "data")),
+        state=dict(already_loaded=State(f"{prefix}-dashboard-loaded", "data")),
+    )
+    def load_default_on_init(collection_name, already_loaded):
+        print(f"[INIT_LOAD] collection_name: {collection_name}, already_loaded: {already_loaded}")
```
J08nY: Logging instead of prints.

sec_certs_page/dashboard/callbacks/dashboard.py
```diff
@@ -0,0 +1,332 @@
...
+        print(f"[SAVE] Total charts to save: {len(charts)}")
+
+        # Load existing dashboard or create new one
+        if dashboard_id:
+            dashboard = dashboard_manager.get_dashboard(dashboard_id)
+            if dashboard and dashboard.user_id == user_id:
+                dashboard.name = dashboard_name or "Untitled Dashboard"
+                dashboard.charts = charts
+            else:
+                # Dashboard not found or doesn't belong to user, create new
```
J08nY: Do we really want to create one if it does not exist? How could it arise that dashboard_id is present yet it does not exist on our end?

I would just error out in this case. If not, the code below is also wrong:
```python
is_new_dashboard = dashboard_id is None
```
while in this branch a new dashboard is created yet it is the `dashboard_id != None` branch.

sec_certs_page/dashboard/chart/base.py
```diff
@@ -0,0 +1,138 @@
+from abc import ABC, abstractmethod
```
J08nY: This directory is missing an `__init__.py` please add it.

sec_certs_page/dashboard/protocols/dashboard.py
```diff
@@ -0,0 +1,46 @@
+import uuid
```
J08nY: This directory is missing an `__init__.py` please add it.

sec_certs_page/dashboard/protocols/chart.py
J08nY: Empty file, why?

sec_certs_page/dashboard/callbacks/utils.py
```diff
@@ -0,0 +1,75 @@
...
+from dash import html
+from dash.development.base_component import Component
+from flask_login import current_user
+
+
+def get_current_user_id() -> str | None:
+    try:
+        if current_user and current_user.is_authenticated:
+            return current_user.id
+    except RuntimeError:
```
J08nY: This seems excessive. If you are getting RuntimeErrors here, then something else is wrong.

sec_certs_page/dashboard/chart/base.py
```diff
@@ -0,0 +1,138 @@
...
+class BaseChart(ABC):
+    """Abstract base for runtime chart components.
+
+    Provides common rendering utilities and defines the interface for chart implementations.
+    Subclasses must implement `title` property and `render` method.
+    """
+
+    graph_id: str
+    data_service: DataService
+    chart_type: Literal["pie", "bar", "box", "line", "scatter", "histogram"]
```
J08nY: There is the `AvailableChartTypes` enum. Why not use that?

sec_certs_page/dashboard/chart/cc/certs_per_year.py
```diff
@@ -0,0 +1,62 @@
...
+
+import plotly.express as px
+from dash.development.base_component import Component
+
+from ...chart.chart import Chart
+from ...data import DataService
+from ..base import BaseChart
+
+
+class CCCertsPerYear(BaseChart):
```
J08nY: Same as for the other custom chart classes.

sec_certs_page/dashboard/chart/cc/validity_duration.py
```diff
@@ -0,0 +1,71 @@
...
+import pandas as pd
+import plotly.express as px
+from dash.development.base_component import Component
+
+from ...chart.chart import Chart
+from ...data import DataService
+from ..base import BaseChart
+
+
+class CCValidityDuration(BaseChart):
```
J08nY: Same as for the other custom chart classes.

sec_certs_page/dashboard/chart/factory.py
```diff
@@ -0,0 +1,135 @@
...
+
+        return chart_class(
+            graph_id=str(config.chart_id),
+            data_service=data_service,
+            chart_type=config.chart_type.value,
+            config=config,
+        )
+
+    @classmethod
+    def create_figure(cls, config: Chart, df: pd.DataFrame) -> go.Figure:
```
J08nY: This is a duplicate method of `FigureBuilder`, why?

sec_certs_page/dashboard/chart/graphs.py
J08nY: Let's rename this file and not use "graphs".

sec_certs_page/dashboard/chart/chart.py
```diff
@@ -0,0 +1,284 @@
...
+    chart_type: AvailableChartTypes
+    x_axis: AxisConfig
+    collection_type: CollectionName
+    title: str = ""
+    order: int = 0
+    y_axis: AxisConfig | None = None
+    color_axis: AxisConfig | None = None
+    filters: dict[str, FilterSpec] = field(default_factory=dict)
+    filter_values: dict[str, Any] = field(default_factory=dict)
+    query_pipeline: list[dict[str, Any]] | None = None
```
J08nY: This `Chart` object gets passed around to the client and back right? (as part of `chart_configs` and such). Then the `query_pipeline` should not be in it. We have no way of validating whatever the client is sending us when it is in a form of an aggregation pipeline.

We basically need to have some internal Chart Config that we then serialize/deserialize in two cases: 
 - To the client. Here we need to be really careful about what we send out and how we validate what we receive back.
 - To the database. Here we can store and load whatever. We trust it.

sec_certs_page/dashboard/pages/fips.py
```diff
@@ -0,0 +1,26 @@
+"""FIPS 140 collection dashboard page."""
```
J08nY: Not sure where to put this comment, but when I open the FIPS "Add chart" modal I get the "Validation Year" in the x-axis list twice:

Image

sec_certs_page/dashboard/filters/query_builder.py
```diff
@@ -0,0 +1,463 @@
...
+from ..filters.filter import FilterSpec
+from ..filters.registry import FilterSpecRegistry, get_all_registries, get_filter_registry
+from ..types.common import CollectionName
+from ..types.filter import AggregationType, FilterOperator
+
+if TYPE_CHECKING:
+    from ..chart.chart import Chart
+
+
+_DERIVED_FIELDS: frozenset[str] = frozenset({"year_from", "year_to"})
```
J08nY: "count" should be in there, perhaps the other aggregation results as well?

sec_certs_page/dashboard/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+
+This ensures the initial HTML payload contains the correct page content,
+eliminating the blank page issue. Client-side routing still works normally
+for subsequent navigation.
+
+Important Notes
+---------------
+- System skips SSR for internal Dash routes (``/_dash-layout``, ``/_dash-dependencies``)
+  as these are AJAX requests for client-side hydration
+- System mutates ``dash.page_container`` singleton, which is safe because:
```
J08nY: This is a weird argument. Yes, in production this will be run with multiple worker processes which will have their own variables and thus a `dash.page_container`. However, mutating any global state like that is just bad. Any state in the app should either be in the DB or be only bound to the request and the request context. No globals should change like this. No matter the arguments about threads or processes.

J08nY: I would look more at https://github.com/plotly/dash/blob/dev/dash/dash.py#L2500 and see what is going wrong there for the page load not working.

sec_certs_page/dashboard/dashboard.py
```diff
@@ -0,0 +1,220 @@
...
+    while allowing users to have multiple dashboards with the same name.
+
+    :param user_id: The user who owns the dashboard
+    :param collection_name: The dataset type (cc, fips, pp)
+    :param created_at: The creation timestamp
+    :return: A UUID5 generated from the namespace and dashboard identifiers
+    """
+    timestamp_str = created_at.isoformat()
+    identifier = f"{user_id}:{collection_name}:{timestamp_str}"
+    return uuid.uuid5(DASHBOARD_NAMESPACE, identifier)
```
J08nY: Why not just random UUID? How is this better?

sec_certs_page/dashboard/data.py
```diff
@@ -0,0 +1,399 @@
...
+
+            df = pd.DataFrame(data)
+            logger.info(f"Aggregation returned {len(df)} records")
+            return df
+
+        except Exception as e:
+            logger.exception(f"Error executing aggregation pipeline on {collection_type.value}")
+            raise e
+
+    def get_distinct_values(self, field: str, dataset_type) -> list[Any]:
```
J08nY: Why does this have a string `dataset_type` while other similar usecases have `CollectionName`?

sec_certs_page/dashboard/data.py
```diff
@@ -0,0 +1,399 @@
...
+                    categories=sorted(unique_eals),
+                    ordered=True,
+                )
+
+        if "not_valid_before" in df.columns:
+            df["year_from"] = pd.DatetimeIndex(df["not_valid_before"]).year
+
+        return df
+
+    def get_dataset_metadata(self, dataset_type: str) -> tuple[dict[str, ColumnStats], int, int]:
```
J08nY: Same as above, str vs `CollectionName`.

sec_certs_page/dashboard/factory.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def __init__(self, collection_name: CollectionName):
+        """
+        Initialize the factory for a specific collection.
+
+        :param collection_name: The collection this factory renders dashboards for
+        """
+        self.collection_name = collection_name
+
+    def render_dashboard_content(
```
J08nY: This whole function is unused. Bad refactor?

sec_certs_page/dashboard/manager.py
```diff
@@ -0,0 +1,219 @@
...
+        if dashboard is None:
+            return None, []
+
+        chart_instances = []
+        for chart_config in dashboard.charts:
+            try:
+                chart_instance = ChartFactory.create_chart(chart_config, self.data_service)
+                chart_instances.append(chart_instance)
+            except ValueError as e:
+                print(f"Warning: Could not instantiate chart {chart_config.name}: {e}")
```
J08nY: Is this reasonable? Should this not just fail fully?

sec_certs_page/dashboard/manager.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        Get predefined chart configurations for 'Load Predefined' option.
+
+        Returns Chart dataclass instances (not BaseChart) for serialization.
+        """
+        registry = self.chart_registries.get(dataset_type)
+        if not registry:
+            return []
+
+        return [chart.config for chart in registry]
```
J08nY: This will also get the custom chart configs if they were previously added to the registry.  Hence my comments about not using that global state.

sec_certs_page/static/dashboard.css
```diff
@@ -0,0 +1,499 @@
...
+.rc-slider-handle:hover,
+.rc-slider-handle:active {
+    border-color: var(--bs-primary) !important;
+    box-shadow: 0 0 0 5px rgba(var(--bs-primary-rgb), 0.2) !important;
+}
+
+/* ============================================
+   Dark theme adjustments
+   ============================================ */
+[data-bs-theme="dark"] .card {
```
J08nY: These I think will mess with our bootstrap styling in general.

J08nY (commented):
I had like 75 comments in this review. Some GitHub UI glitch is only showing 40 now. Lets hope it is only a glitch.

pyproject.toml
xvalec01: Okay.

sec_certs_page/__init__.py
```diff
@@ -220,7 +217,54 @@ def get_searcher():
...
+
+        from .common.permissions import dashboard_permission
+
+        @app.before_request
+        def check_dashboard_access():
+            if request.path.startswith(url_base_pathname):
+                if not current_user.is_authenticated:
+                    return redirect(url_for("user.login", next=request.url))
+                if not dashboard_permission.can():
+                    from flask import abort
```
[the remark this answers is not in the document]
xvalec01: Done.

sec_certs_page/dashboard/pages/common.py
```diff
@@ -0,0 +1,712 @@
...
+import dash_bootstrap_components as dbc
+from dash import dcc, html
+
+from ... import DASHBOARD_URL_BASE_PATHNAME
+from ..types.chart import AvailableChartTypes
+from ..types.common import CollectionName
+from ..types.filter import AggregationType
+
+# Data type to available aggregations mapping
+DATA_TYPE_AGGREGATIONS: dict[str, list[AggregationType]] = {
```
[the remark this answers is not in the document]
xvalec01: Yes, you are right. I don't remember why I put it here.

sec_certs_page/dashboard/__init__.py
xvalec01: I don't understand, I spent like 2 days on it, went deep. The problem was with the callback inside the `router` method. The `prevent_initial_call=True` and `Output(_ID_CONTENT, "children")` caused that when accessing the `localhost:5000/dashboard/` this callback emptied the children after accessing the page, so that resulted into children being empty. I removed the code and it is not doing that anymore. So something that I changed was probably causing it and this change is needed no more. I think I am not able to replicate it.

xvalec01:
I will finish going through comments on Saturday.

sec_certs_page/dashboard/callbacks/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    from ..manager import DashboardManager
+
+
+def register_all_callbacks(
+    dash_app: "Dash",
+    data_service: "DataService",
+    filter_factories: dict[CollectionName, FilterFactory],
+    chart_registries: dict[CollectionName, "ChartRegistry"],
+    dashboard_manager: "DashboardManager",
+) -> None:
```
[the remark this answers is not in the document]
xvalec01: Done.

sec_certs_page/dashboard/callbacks/charts.py
```diff
@@ -0,0 +1,232 @@
...
+
+        triggered_id = ctx.triggered_id
+
+        if triggered_id == f"{prefix}-add-chart-btn":
+            if selected_chart_id and selected_chart_id not in current_configs:
+                # Add the chart config (for predefined charts, get from registry)
+                predefined = chart_registry.get_predefined(selected_chart_id)
+                if predefined and predefined.config:
+                    current_configs[selected_chart_id] = predefined.config.to_dict()
+                    print(f"[CHART_MGMT] Added predefined chart {selected_chart_id} with config")
```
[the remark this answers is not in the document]
xvalec01: Should I log in the dashboard system now in some way and you will just edit it later with logger setup?

sec_certs_page/dashboard/callbacks/charts.py
```diff
@@ -0,0 +1,232 @@
...
+                ]
+            )
+
+        rendered = []
+
+        for chart_id in chart_ids:
+            chart = chart_registry.get(chart_id)
+
+            # If not in registry, try to create from chart_configs store
+            # Register in active charts for caching during this session
```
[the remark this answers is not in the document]
xvalec01: So this could happen when user lost the server side in-memory data so it takes the data from client-side storage. And if neither of that if you saved the dashboard, these representations of Dashbooard, Charts and Filters are in the monog db dashboard collection.

sec_certs_page/dashboard/dashboard.py
xvalec01: I probably ruined the purpose how it was intended with timestamp. It should for specific settings of the dashboard always create the same UUID.

sec_certs_page/dashboard/protocols/filter.py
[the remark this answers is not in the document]
xvalec01: Same as previously

sec_certs_page/dashboard/protocols/dashboard.py
xvalec01: Will remove the protocols directory, no need.

sec_certs_page/dashboard/callbacks/dashboard.py
xvalec01: Changes to incorporate this in this commit ae86af30

sec_certs_page/dashboard/callbacks/charts.py
```diff
@@ -0,0 +1,232 @@
...
+                            f"Chart '{chart_id}' not found.",
+                        ],
+                        color="warning",
+                    )
+                )
+                continue
+
+            try:
+                chart_component = chart.render(filter_values or {})
+                is_editable = chart.config.name.startswith("custom-") if chart.config else False
```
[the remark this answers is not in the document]
xvalec01: Yeah, you are right. Improved in e53f7830

sec_certs_page/dashboard/callbacks/utils.py
xvalec01: Yes. I left it like that accidentally. 8b059bae

sec_certs_page/dashboard/callbacks/filters.py
```diff
@@ -0,0 +1,132 @@
...
+            filter_specs=Output(f"{prefix}-filter-specs", "data"),
+            metadata_loaded=Output(f"{prefix}-metadata-loaded", "data"),
+        ),
+        inputs=dict(modal_open=Input(f"{prefix}-create-chart-modal", "is_open")),
+        state=dict(already_loaded=State(f"{prefix}-metadata-loaded", "data")),
+        prevent_initial_call=True,
+    )
+    def load_metadata(modal_open, already_loaded):
+        """Load metadata only when modal opens for the first time - lazy loading optimization."""
+        from dash import no_update
```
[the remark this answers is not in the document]
xvalec01: No need. Removed. d7716e9a

sec_certs_page/dashboard/callbacks/modal.py
```diff
@@ -0,0 +1,852 @@
...
+            if i < len(modal_filter_values) and modal_filter_values[i]:
+                value = modal_filter_values[i]
+                if value != "__all__":
+                    chart_filter_values[spec["id"]] = value
+
+    combined_filters = {**(current_filter_values or {}), **chart_filter_values}
+
+    chart_config = Chart(
+        chart_id=chart_id,
+        name=f"custom-{chart_id}",
```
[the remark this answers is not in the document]
xvalec01: Yes. Changed. bdfc0e99

sec_certs_page/dashboard/chart/registry.py
```diff
@@ -0,0 +1,63 @@
...
+        """Remove a chart from the active registry.
+
+        :param chart_id: The ID of the chart to unregister
+        """
+        if chart_id in self._active_charts:
+            del self._active_charts[chart_id]
+
+    def clear_active(self) -> None:
+        """Clear all active charts (but keep predefined ones)."""
+        self._active_charts.clear()
```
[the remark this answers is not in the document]
xvalec01: Yeah, you are right. That needs to be reworked. I didn't realize it. I had this class since the beggining of the implementation and just kept it without questioning it. I will implement Redis to the system.

sec_certs_page/dashboard/chart/cc/category_distribution.py
```diff
@@ -0,0 +1,59 @@
...
+
+import plotly.express as px
+from dash.development.base_component import Component
+
+from ...chart.chart import Chart
+from ...data import DataService
+from ..base import BaseChart
+
+
+class CCCategoryDistribution(BaseChart):
```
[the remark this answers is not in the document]
xvalec01: Yes, I will rework it. But I wanted a class because some of the operations with filters and config could not be done so I used Pandas to get the specific data. I can now specify the predefined values for query builder and use mongo for the calculation as it is with the custom charts.

sec_certs_page/dashboard/chart/graphs.py
```diff
@@ -0,0 +1,126 @@
...
+from .base import BaseChart
+from .figure_builder import FigureBuilder
+
+
+class GenericChartComponent(BaseChart):
+    """Base for generic chart components that render from configuration.
+
+    Unlike predefined charts (CCCertsPerYear, etc.) that have hardcoded
+    data transformations, generic charts use the Chart configuration's
+    x_axis/y_axis/aggregation settings to dynamically create visualizations.
```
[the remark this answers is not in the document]
xvalec01: Predefined charts removed and now they use the same approach as the generic ones for chart building from UI.

sec_certs_page/dashboard/chart/factory.py
```diff
@@ -0,0 +1,135 @@
...
+    LineChartComponent,
+    PieChartComponent,
+    ScatterChartComponent,
+)
+from ..data import DataService
+from ..types.chart import AvailableChartTypes
+from ..types.filter import AggregationType
+
+
+class ChartFactory:
```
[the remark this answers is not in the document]
xvalec01: I think this is now fixed.

sec_certs_page/dashboard/chart/chart.py
xvalec01: Resolved here in commit 4c7411c0

sec_certs_page/dashboard/callbacks/charts.py
```diff
@@ -0,0 +1,232 @@
...
+            )
+
+        rendered = []
+
+        for chart_id in chart_ids:
+            chart = chart_registry.get(chart_id)
+
+            # If not in registry, try to create from chart_configs store
+            # Register in active charts for caching during this session
+            if not chart and chart_configs and chart_id in chart_configs:
```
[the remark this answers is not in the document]
xvalec01: I changed it as you requested and then changed the code quite a lot because of other problems that you found.

sec_certs_page/dashboard/types/chart.py
```diff
@@ -0,0 +1,32 @@
+from datetime import datetime
+from enum import Enum
+from typing import Any, TypedDict
+from uuid import UUID
+
+
+class AvailableChartTypes(str, Enum):
```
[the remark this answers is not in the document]
xvalec01: In Python 3.11, there is StrEnum. Until then, it is common to use str and Enum inheritance.

sec_certs_page/dashboard/types/chart.py
```diff
@@ -0,0 +1,32 @@
...
+    BAR = "bar"
+    STACKED_BAR = "stacked_bar"
+    LINE = "line"
+    PIE = "pie"
+    SCATTER = "scatter"
+    BOX = "box"
+    HISTOGRAM = "histogram"
+
+
+class ChartDict(TypedDict):
```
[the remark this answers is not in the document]
xvalec01: Removed

PR #536, opened 2025-11-30
Preparing for EUCC: Moving heuristics functions to a common file

tkachyna:
One of the first pull requests which will involve moving shared functionality (for both CC and EUCC) into common files

These pull requests will be structured according to the individual parts of the pipeline. This one includes heuristic functions.


adamjanovsky:
@tkachyna sorry, I didn't make the promised review today. Making it a top priority for tomorrow morning.

tkachyna:
> Both assumptions get broken with new functionality and I'm wondering how to address this. I guess we still have types that help us (the method acception CCCertificate | EUCCCertificate, for example). But is there any additional pattern we could maintain to help this distinction?

The easiest way I can think of is to give the methods a prefix (like eucc, cc, fips etc.), but it's definitely something we can discuss at the meeting on Thursday.

PR #537, opened 2025-11-30
Br1 extraction

ndavidova:


src/sec_certs/br1/chapter_parsing/mapper.py
```diff
@@ -0,0 +1,82 @@
...
+    if subchapter_num > 0:
+        title = base_chapter.subchapters[subchapter_num - 1].title
+
+    title = substitute(title)
+
+    return rf"^(##|Section)*{chapter_num}(\.?{subchapter_num})?\.?{title}$"
+
+
+# Core extraction logic
+def extract_chapters_from_text(text: str) -> list[Chapter]:
```
J08nY: As I understand it, this uses the Markdown output of Docling to parse out the chapter. Thus, it introduces additional dependencies on Markdown parsing and does a somewhat unnecessary round trip where Docling parses the PDF to some internal format (which it dumps as JSON) and then serializes it to Markdown which is parsed here. I think this is an OK solution for now. But when #525 lands, we should consider parsing out the chapters directly from that JSON.

src/sec_certs/br1/table_parsing/model/entry_types/algorithms.py
```diff
@@ -0,0 +1,34 @@
+from dataclasses import dataclass
+
+# Tables 5,6,7,8,9
+# AlgoProp is actually a Key:Value pair, but for simplicity it is represented as a string
+
+
+@dataclass
+class ApprovedAlgo:
```
J08nY: All of these DataClasses are quite verbose, though I like that it allows the type information to carry a lot of the structure and meta information about the data. Do you have any additional descriptions of these fields from the BR1? If so, could you (perhaps in an automated way or not) add them as docstrings to the classes and attributes? Just a single sentence (if available) would suffice.

The names are usually quite descriptive so this is not that important, more of a nice to have.

src/sec_certs/sample/fips.py
```diff
@@ -629,6 +637,26 @@ def extract_policy_pdf_keywords(cert: FIPSCertificate) -> FIPSCertificate:
             cert.pdf_data.keywords = keywords
         return cert
 
+    @staticmethod
+    def extract_br1_metadata(cert: FIPSCertificate) -> FIPSCertificate:
+        """
+        Extract br1 chapters and tables from the document
+        """
+        with Path(cert.state.policy_txt_path).open() as f:
```
J08nY: When #525 lands, this should perhaps be gated behind checking that the PDFConverter is the Docling one. Because if it is not, it does not even make sense to run all of this, because it will not produce the proper chapters right?

J08nY (commented):
Cool PR. Thanks for the work! Really nicely done. It adds a bunch of code, though I guess that is understandable and most of it is in the JSON describing the format and the table dataclasses. I am not sure whether `sec_certs.br1` is the right package for this. I think it makes the most sense to have it under `sec_certs.br1.heuristics`? (wdyt @adamjanovsky ?) Even though it is not a heuristic necessarily (the matching is a bit of a heuristic). Just having it top-level feels weird given that it is a bit of a niche feature in the grand scheme of things.

Also, when #525 lands, there should be tests that show the *successful* parsing of a BR1 policy.

PR #540, opened 2025-12-09
refactor: Dashboard layout and styles

xvalec01:


J08nY:
I played with this a bit and adding the pre-defined charts for CC was not working for me, the charts were not correct.

Furthermore, when I navigated from the Dashboard home to the CC or FIPS home I got a 404 as the link pointed to some `/CollectionName` page instead of `/cc` or `/fips`.

J08nY:
Is this review ready?

PR #549, opened 2026-01-04
refactor: add CVSS v4.0 support to CVE sample

### LINKED ISSUE #532: Parse and use CVSS 4.0 metrics
This is a follow-up on #529.

We should:
 - Parse CVSS 4.0 metrics and extract useful info.
 - Perhaps remove the exploitabilityScore and impactScore from the CVE dataset if we cannot reliably extract that for CVSS 4.0.

adamjanovsky:
Closes #532 

- Rename Metrics to CVSS and add version field
- Remove exploitability_score and impact_score (not in CVSS v4.0)
- Support parsing cvssMetricV40 from NVD API
- Update test data and fixtures


J08nY:
> Following a good tradition, this is a breaking change, so new datasets must be re-generated and served by sec-certs.org.
> 
> @J08nY could you briefly review this?

What do you mean by re-generated? Every week the run is fresh and then gets diffed and merged with the database. Is anything else necessary?

adamjanovsky:
> > Following a good tradition, this is a breaking change, so new datasets must be re-generated and served by sec-certs.org.
> > @J08nY could you briefly review this?
> 
> What do you mean by re-generated? Every week the run is fresh and then gets diffed and merged with the database. Is anything else necessary?

note, that will suffice.

PR #550, opened 2026-01-07
Preparing for EUCC: Moving convert and extraction functions to common files 

tkachyna:
Second pull requests which will involve moving shared functionality (for both CC and EUCC) into common files.

These pull requests will be structured according to the individual parts of the pipeline. This one includes convert and extractions functions.

! I had a problem with multiprocessing. After moving convert_report_pdf, convert_st_pdf, convert_cert_pdf functions outside of the CCCertificate class to a common file, a function which handles multiprocession (convert_pdfs) had a problem with these functions not being static anymore. The easiest solution was currently to move them to  dataset/common.py from sample/common.py ! Maybe I will find easier solution in the upcoming weeks.

### DIFF
--- a/src/sec_certs/dataset/cc.py
+++ b/src/sec_certs/dataset/cc.py
@@ -25,7 +25,9 @@ from sec_certs.dataset.auxiliary_dataset_handling import (
 )
 from sec_certs.dataset.common import (
     compute_heuristics_body,
-    convert_all_pdfs_body,
+    convert_certs_pdfs,
+    convert_reports_pdfs,
+    convert_targets_pdfs,
     download_all_artifacts_body,
     extract_all_frontpages,
     extract_all_keywords,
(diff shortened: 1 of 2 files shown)


adamjanovsky (approved):
Hi,

the problem was that the `sample/common.py` did: `from sec_certs.dataset.dataset import logger` which created a circular import. I have no idea how the CI tests passed with the current code.

Anyway, I implemented the fix (importing the logger from `sec_certs.sample.certificate`) and moved the `convert_report_pdf()`, `convert_st_pdf()`, `convert_cert_pdf()` methods back to `sample/common.py` where they naturally belong. 

I also merged fresh main to this. 

I think the `dataset/common.py`

```python
def download_all_artifacts_body(obj, fresh: bool = True) -> None:
    download_reports(obj, fresh)
    download_targets(obj, fresh)
    download_certs(obj, fresh)
```

is not very descriptive. It's not clear which datasets this concerns. I'd recommend having the following in `dataset/cc.py`

```python
def _convert_all_pdfs_body(self, converter_cls: type[PDFConverter], fresh: bool = True) -> None:
    download_reports(obj, fresh)
    download_targets(obj, fresh)
    download_certs(obj, fresh)
```

Also, I'll rework this logic quite substantially soon. I'll get in touch to coordinate.

Issue #547, opened 2026-01-04
Failing docling tests due to different outputs


adamjanovsky:
Two tests fail on my MacOS:

```
FAILED tests/cc/test_cc_dataset.py::test_convert_pdfs[DoclingConverter] - AssertionError: assert 7174 < 1000
FAILED tests/fips/test_fips_dataset.py::test_convert_pdfs[DoclingConverter] - assert 1043 < 1000
```

These tests compare produced text files against templates. See the CC actually produced by MacOS attached: [test.md](https://github.com/user-attachments/files/24421401/test.md). There are some mild differences compared to template, but nothing really serious.  Lot of it is caused by whitespace. 

I suggest not investigating the root cause: could be PDF parsing, could be [OCR actually taking over](https://github.com/docling-project/docling/issues/2312) and using different engine), could be something else. Instead, we could just relax these assertions in favour of some sanity checks on the content (heading actually present, some data here and there). 

@jborsky what do you think?

PR #551, opened 2026-01-18
Fix failing docling tests due to different outputs

jborsky:
This PR fixes https://github.com/crocs-muni/sec-certs/issues/547.

I've decided to relax the assertions using the insertion/deletion similarity between the template and the actual conversion result. 

First, it normalizes the whitespaces and then computes the similarity using the RapidFuzz library, which I've leveraged since it is already in the project deps.

The threshold was set by playing around with the templates and actual content on my MacBook, and can be adjusted as needed.


adamjanovsky (approved):
LGTM, tests pass 🎉 

  ```
= 176 passed, 23 skipped, 38 deselected, 1 xfailed, 2 xpassed, 26 warnings in 416.56s (0:06:56) =
```

PR #552, opened 2026-01-21
Add fetching and processing for EUCC Certificates + refactoring the rest of the pipeline

tkachyna:
This pull request adds:
- logic for scraping the official ENISA webpage, parsing the metadata obtained from this page and initializing the EUCC dataset and certificates (`src/sec-certs/dataset/eucc.py`, `src/sec-certs/sample/eucc.py` )
- refactors common logic shared between EUCC and CC into files named 'cc_eucc_common.py'.
Since CC and EUCC share a significantly larger portion of code compared to other datasets in the code base, using this prefix makes the intent explicit and immediately signals to developers that this logic is meant only for these two datasets.
Type safety is still preserved, ensuring that no other dataset or certificate types can be passed to the functions defined in these common files.
    -  originally I wanted to solve these common files problems by defining a Mixin class, which CCDataset and EUCCDataset would implement. However I ran into issues with static analysis and mypy, because mypy could not see references to fields used in the Mixin that were defined in the Dataset class. For some reason, I thought this would work as I was assuming Python has something similar to Scala self-types which I work with everyday, but unfortunately, it does not
- new regex for german eucc certificates
- tests for the new functions will be added in one of the other PR

### DIFF
--- a/src/sec_certs/dataset/cc.py
+++ b/src/sec_certs/dataset/cc.py
@@ -14,7 +14,6 @@ from pydantic import AnyHttpUrl
 
 from sec_certs import constants
 from sec_certs.configuration import config
-from sec_certs.converter import PDFConverter
 from sec_certs.dataset.auxiliary_dataset_handling import (
     AuxiliaryDatasetHandler,
     CCMaintenanceUpdateDatasetHandler,


src/sec_certs/dataset/cc.py
```diff
@@ -155,6 +151,10 @@ def to_pandas(self) -> pd.DataFrame:
 
         return df
 
+    @property
+    def dataset_name(self) -> str:
+        return "CC"
+
```
adamjanovsky: The `Dataset` class already defines `name` attribute, see: https://github.com/crocs-muni/sec-certs/blob/6376b6968ae233b2f3ea2c22b60f3b581987faee/src/sec_certs/dataset/dataset.py#L86

CC and EUCC classes should preferrably work with that attribute.

src/sec_certs/utils/helpers.py
```diff
@@ -348,3 +349,13 @@ def choose_lowest_eal(eals: set[str] | None) -> str | None:
...
         return eal_entries[0][2]
     return None
+
+
+class DocType(Enum):
+    REPORT = ("report", "certification report")
+    TARGET = ("st", "security target")
+    CERTIFICATE = ("cert", "certificate")
+
+    def __init__(self, short: str, long: str):
```
adamjanovsky: Is this necessary? I didn't spot any invocation of this through the codebase.

src/sec_certs/sample/eucc.py
```diff
@@ -0,0 +1,213 @@
...
+            return {}
+
+        dgst = helpers.get_first_16_bytes_sha256(certificate_id)
+
+        product_type = metadata.get("product_type", "").upper()
+        product_name = metadata.get("product_name", "")
+        holder_name = metadata.get("holder_name", "")
+        scheme = EUCCCertificate._get_scheme_from_cert_id(certificate_id)
+        security_level = metadata.get("package", "")
+        not_valid_before = EUCCCertificate._get_not_valid_before(metadata.get("issuance_date_full"))
```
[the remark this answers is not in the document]
adamjanovsky: Ok. If they're recovered from artifacts, they should not be stored within `EUCCCertificate.not_valid_before``/after` fields. They would naturally belong to heuristics, but new attribute is not warranted by 2 cases. I suggest ignoring the computation and filling the `not_valid_before` attributes with None. 

This is how the current PR behaves, right?

src/sec_certs/dataset/eucc.py
```diff
@@ -0,0 +1,437 @@
...
+                CPEDatasetHandler: CPEDatasetHandler(self.auxiliary_datasets_dir if self.is_backed else None),
+                CVEDatasetHandler: CVEDatasetHandler(self.auxiliary_datasets_dir if self.is_backed else None),
+                CPEMatchDictHandler: CPEMatchDictHandler(self.auxiliary_datasets_dir if self.is_backed else None),
+                CCSchemeDatasetHandler: CCSchemeDatasetHandler(self.auxiliary_datasets_dir if self.is_backed else None),
+                ProtectionProfileDatasetHandler: ProtectionProfileDatasetHandler(
+                    self.auxiliary_datasets_dir if self.is_backed else None
+                ),
+            }
+
+    @property
```
adamjanovsky: Similar note as to cc dataset name applies.

adamjanovsky (changes requested):
Thanks for the PR 👍 . Few notes:

- I'd say it's generally good idea to have PR and the corresponding tests in a single PR. If the tests uncover some problems, you'll be addressing them in somewhat unrelated PR which is not ideal. But proceed as you wish in this case.
- I agree that inheritance or mix-ins are not ideal in this case. The current approach is somewhat wordy but fairly easy to understand. 
- I requested some minor changes here and there.

adamjanovsky (approved):

PR #565, opened 2026-03-16
New MIP statuses

jborsky:
- Added new MIP statuses. 
- In computing MIPFlows from snapshots, if one snapshot contained duplicate entries, they were sorted and the first was picked. I slightly adjusted the logic. For newer snapshots, it now uses the most recent status_since field. Tho it seems it's not an issue for newer snapshots? For older snapshots, it falls back to using the status, but now selects the most progressed status instead. It makes more sense to me like that. Am I missing some context?

### DIFF
--- a/src/sec_certs/sample/fips_mip.py
+++ b/src/sec_certs/sample/fips_mip.py
@@ -53,6 +53,7 @@ class MIPEntry(ComplexSerializableType):
     standard: str
     status: MIPStatus
     status_since: date | None
+    submission_id: str | None = None
 
     def to_dict(self) -> dict[str, str | MIPStatus | date | None]:
         return {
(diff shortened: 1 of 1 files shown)


jborsky:
> Huh, did you not need to change the snapshot download function? Interesting.

No, neither the site nor the table with entries changed except for the new statuses.

> There are many duplicates in the list, sometimes with different statuses or dates. I am not sure what the meaning of that is and how to correctly handle that, or how we even handle it. 

Duh, how I could have missed it... You are right, it's the case in the newer snapshots as well. 

On the page, we are not handling it. Just when listing status changes, only one of the duplicate entries is selected to compute the days and show the status changes. As you said, we need to figure out the semantics of the duplicates to be able to handle them better.

PR #610, opened 2026-03-20
Accept CVEs with status: 'Deferred' into the dataset

### LINKED ISSUE #609: NVD deferred entries missing from dataset
With NVD resource constraints leading to introduction of Deferred state, we've lost visibility into many CVEs that are actually populated with CPEs.

See https://www.nist.gov/itl/nvd

> **April 2, 2025: NVD General Announcement**
> (Note: this statement was updated on April 10, 2025 to clarify which CVEs will be deferred.)
> All CVEs with a published date prior to 01/01/2018 that are awaiting further enrichment will be marked as Deferred within the NVD dataset.
> We are assigning this status to older CVEs to indicate that we do not plan to prioritize updating their enrichment data due to the CVE’s age.

This affects perfectly well populated CVEs, e.g., RoCA: https://nvd.nist.gov/vuln/detail/cve-2017-15361. As a result, they go missing from our dataset due to the following if:

https://github.com/crocs-muni/sec-certs/blob/b50a95aecc56779fe2ea0bd6b525d673170f5f0b/src/sec_certs/dataset/cve.py#L183-L184

The main motivation to introduce the if was to lower the dataset size by discarding useless entries. Given that many `Deferred` CVEs are pretty useful, we should also accept `Deferred` CVEs into the dataset.

adamjanovsky:
Closes #609

### DIFF
--- a/src/sec_certs/dataset/cve.py
+++ b/src/sec_certs/dataset/cve.py
@@ -180,7 +180,7 @@ class CVEDataset(JSONPathDataset, ComplexSerializableType):
         self.last_update_timestamp = datetime.fromisoformat(data["timestamp"])
         for vuln in data["vulnerabilities"]:
             # https://nvd.nist.gov/vuln/vulnerability-status#divNvdStatus
-            if vuln["cve"]["vulnStatus"] in {"Analyzed", "Modified"}:
+            if vuln["cve"]["vulnStatus"] in {"Analyzed", "Modified", "Deferred"}:
                 cve = CVE.from_nist_dict(vuln["cve"])
                 self[cve.cve_id] = cve
         return self


jborsky (approved):

J08nY:
@adamjanovsky What about the other CVE states? I think we should not filter on the CVEDataset level and that it should contain all CVEs and we should only filter on-demand when we want to do the matching.

jborsky:
I'm not definitely against making the CVE dataset more general since it's also one of the features on the list that we put together. I just want to add that I think we currently don't have enough users or feedback data to determine which use cases are genuinely valuable or what our users would expect from our features.

I believe our ideas and existing features are useful and have real potential, but we need to validate whether they are actually needed by potential users in practice. Currently, the main goal from a CCAT perspective is to get as many users as possible and gather as much feedback as possible, which I hope the planned testing after the first dev cycle from different companies will help us with.

J08nY:
Note the current NIST stance:
https://www.aikido.dev/blog/nist-nvd-changes-2026
https://www.nist.gov/news-events/news/2026/04/nist-updates-nvd-operations-address-record-cve-growth

They also updated the statuses:
https://nvd.nist.gov/vuln/vulnerability-status

@adamjanovsky You better understand how this works, can you give like 5 minutes and check that we will be okay with these changes?

adamjanovsky:
The range of [vulnerability statuses](https://nvd.nist.gov/vuln/vulnerability-status) we should consider remains the same, so no code changes are needed.

However, it means that new CVEs likely won't be enriched with CPEs. Our vulnerability analysis will thus with time obsolete. Fixing this would require a lot of effort. I also think that with the advent of LLMs, the preferred matching approach should change (from strict CPEs to conext-aware LLM decisions). 

I think the default is now "do nothing" and mark the cert <-> CVE mapping fixes as future work.

PR #629, opened 2026-04-14
Introduce DocumentState for FIPSCertificate class

### LINKED ISSUE #464: Introduce `DocumentState` for `FIPSCertificate` class

adamjanovsky:
Closes #464

### DIFF
--- a/src/sec_certs/dataset/fips.py
+++ b/src/sec_certs/dataset/fips.py
@@ -126,7 +126,7 @@ class FIPSDataset(Dataset[FIPSCertificate], ComplexSerializableType):
         Extracts data from html module file
         """
         logger.info("Extracting data from html modules.")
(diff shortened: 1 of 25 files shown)


jborsky:
Looks good to me, thanks. 

These changes would break the page tho, so it makes sense to address them in another PR before this one lands to ensure compatibility. I'll handle that.

jborsky (approved):

J08nY:
A note for the web: This will likely need fixing in the comparison: https://github.com/crocs-muni/sec-certs/blob/page/sec_certs_page/common/diffs.py#L750 The `fips_diff_method` should get the state.

PR #631, opened 2026-04-15
Page compatibility with unified DocumentState

jborsky:
This PR ensures page compatibility with https://github.com/crocs-muni/sec-certs/pull/629, which introduces a unified `DocumentState` across all certificate types.

- Replace all usages of `pdf_path/pdf_hash` with `source_path/source_hash`
- Use the new nested `DocumentState` structure for FIPS, replacing the `module_*/policy_*` fields
- Update diff schemas to reflect the new field names and structure
- Update test fixtures to reflect the new field names and structure, including `json_hash` that was introduced some time ago

### DIFF
--- a/sec_certs_page/cc/tasks.py
+++ b/sec_certs_page/cc/tasks.py
@@ -247,10 +247,10 @@ class CCUpdater(Updater, CCMixin):  # pragma: no cover
 
             with sentry_sdk.start_span(op="cc.move", name="Move files"), suppress_child_spans():
                 for cert in dset:
-                    if cert.state.report.pdf_path and cert.state.report.pdf_path.exists():
+                    if cert.state.report.source_path and cert.state.report.source_path.exists():
                         dst = paths["report_pdf"] / f"{cert.dgst}.pdf"
-                        if not dst.exists() or get_sha256_filepath(dst) != cert.state.report.pdf_hash:
-                            cert.state.report.pdf_path.replace(dst)
+                        if not dst.exists() or get_sha256_filepath(dst) != cert.state.report.source_hash:
+                            cert.state.report.source_path.replace(dst)
                     if cert.state.report.txt_path and cert.state.report.txt_path.exists():
                         dst = paths["report_txt"] / f"{cert.dgst}.txt"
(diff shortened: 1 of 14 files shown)


jborsky:
The `test_load_cert` fails because the main branch lacks updated classes. Once https://github.com/crocs-muni/sec-certs/pull/629 is merged, they will pass.

jborsky:
I've tested processing with the #629 branch and everything looks good, but there's one edge case that needs to be addressed.

When updating the DB with a newly processed dataset, removed certificates are only marked with a removal diff entry; the certificate entry in the DB itself is kept as-is. This means its schema is never updated, which breaks the assumption that all documents conform to the new schema (e.g. the CC entry page breaks for these entries).

I would handle this with a one-time migration step run after deploying the updated library and page on the server. Manually (via a script) update the remaining certificates that were neither newly inserted nor updated by processing, to conform to the new schema (rename the relevant fields, and additionally for FIPS use the new nested DocumentState).

Wdyt @J08nY, would you handle it differently?

J08nY:
Yes this is an issue. I think I encountered it before and solved it the same way. It is unfortunate, but we do not have database migrations in mind, as the format is defined in the tool, which just does not care about the format. For now, I would fix it the way you describe, but perhaps make an issue so we can discuss different approaches in the future.

The problem is broader:
 - The JSON output format is very tightly mapped to tool data structures. This is nice, because there is not a lot of parsing. However, it also means it changes when the tool data structures change.
 - There is no intention of backwards or forward compatibility in the tools JSON output.
 - This problem manifests not only on the website, but also for users that have our datasets processed or downloaded before a format change.
 - Similarly, when the tool updates, but the dataset on the site does not, even simple features like `CCDataset.from_web()` break. The same happens when the user has an older version of the tool and the dataset updated already.

Some of the format changes are clearly inevitable, and for some there is clearly no way to make them fully compatible: there is simply no information in the previous version to create the data for the next version. However, I believe with a bit more careful design around de/serialization we could avoid changing the format so often and could make changes with more backwards/forwards compatibility.

jborsky:
Yes, I agree with your points and the limitations of the current design. I will handle this for now and create the issue for the future.

PR #659, opened 2026-06-12
2.1. Scrape data about certified PPs from NIAP portal

### LINKED ISSUE #578: 2.1. Scrape data about certified PPs from NIAP portal
**Description**
* Investigate the reason of CC portal missing PPs certified and available on NIAP portal
* Extend the PPs dataset with PPs from NIAP

A user wants to find protection profiles that have been most recently certified. There are items that are listed on the NIAP portal but not in sec-certs. Example: Searching "file encryption" in sec-certs brought old PP from 2019 but not the one from 2025 that is up on https://www.niap-ccevs.org/protectionprofiles --> PP-module for File Encryption Version 2.0 (approved 09/29/2025). It is not clear for users why this item does not appear in sec-certs and where sec-certs gets its content from.
* Need to investigate the reason CC portal does not contains some of the certified PPs from NIAP (and potentially other resources)

**Additional context**
CCAT project

m-mikulica:
Adds scraping of US Protection Profiles from the NIAP public API and merges them
into main ProtectionProfileDataset  following the CC scheme-data pattern.
Belongs to issue #578 

- NIAPScraper scrapes the NIAP API:
 -version is parsed from "pp_short_name", status is derived from "sunset_date"
 -"is_collaborative" from the PP name and predecessor/successor from the per-PP
  detail api endpoint
- scraped data is built as an auxiliary dataset("PPSchemeDataset" + "PPSchemeDatasetHandler" )
- scraping runs in "process_auxiliary_datasets"
- matching + enrichment in "compute_heuristics" (same pattern as CC)
- "PPSchemeMatcher" matches scraped records to existing CC-portal PPs
- "ProtectionProfile.heuristics.scheme_data" carries a per-PP enrichment dict
- Tests in `tests/cc/test_pp_scheme.py` covering parsing, mapping, matcher
  enrichment and the handler (+ one remote  test against live NIAP)


src/sec_certs/dataset/pp_scheme.py
```diff
@@ -0,0 +1,66 @@
...
+from collections.abc import Mapping
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from sec_certs.dataset.json_path_dataset import JSONPathDataset
+from sec_certs.sample.pp_scheme import PP_SCHEME_SCRAPERS, PPSchemeRecord
+from sec_certs.serialization.json import ComplexSerializableType
+
+if TYPE_CHECKING:
+    from sec_certs.sample.pp_scheme import PPScraper
```
jborsky: `PPScraper` is not heavy to import, so there is no need to guard it.

src/sec_certs/dataset/pp_scheme.py
```diff
@@ -0,0 +1,66 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class PPSchemeDataset(JSONPathDataset, ComplexSerializableType):
+    """
+    A dataset of Protection Profile records scraped from national scheme portals
+    list of SchemeRecord objects
```
jborsky: The docstring is probably a broken fragment and names a non-existent type.

src/sec_certs/dataset/pp_scheme.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __iter__(self):
+        for records in self.schemes.values():
+            yield from records
+
+    def __len__(self) -> int:
+        return sum(len(v) for v in self.schemes.values())
+
+    def __getitem__(self, scheme: str) -> list[PPSchemeRecord]:
+        return self.schemes[scheme.upper()]
```
jborsky: When populating the schemes in `from_scrapers`, the scheme is not uppercased; it is taken directly from `scraper.scheme`. So, if the scraper defines a lowercase scheme key, this `__getitem__` wouldn't match.  

Remove it or normalize on insert too.

src/sec_certs/dataset/protection_profile.py
```diff
@@ -131,6 +137,14 @@ def web_dir(self) -> Path:
         """
         return self.root_dir / "web"
 
+    @property
+    @only_backed(throw=False)
+    def scheme_data_path(self) -> Path:
```
jborsky: This is unused; the path is defined in the `PPSchemeDatasetHandler`.

src/sec_certs/dataset/protection_profile.py
```diff
@@ -417,21 +431,57 @@ def _extract_pp_keywords(self):
...
-    def process_auxiliary_datasets(self, **kwargs) -> None:
+    def process_auxiliary_datasets(self, download_fresh: bool = False, **kwargs) -> None:
         """
-        Dummy method to adhere to `Dataset` interface. `ProtectionProfile` dataset has currently no auxiliary datasets.
-        This will just set the state `auxiliary_datasets_processed = True`
+        Builds the PP scheme auxiliary dataset by scraping national scheme portals via
+        PPSchemeDatasetHandler. Matching and enrichment is in compute_heuristics.
         """
-        logger.info("Protection Profile dataset has no auxiliary datasets to process, skipping.")
+        self.aux_handlers[PPSchemeDatasetHandler].process_dataset(download_fresh)
```
jborsky: This assumes the`PPSchemeDatasetHandler` couldn't be `None`.

src/sec_certs/dataset/protection_profile.py
```diff
@@ -53,6 +55,10 @@ def __init__(
         aux_handlers: dict[type[AuxiliaryDatasetHandler], AuxiliaryDatasetHandler] | None = None,
```
jborsky, line 55: I know this was not in the scope of your changes, but related to my comment below about the wrong assumption that the `PPSchemeDatasetHandler` couldn't be `None`, a correct type annotation here would catch it.


```suggestion
        aux_handlers: dict[type[AuxiliaryDatasetHandler], AuxiliaryDatasetHandler | None] | None = None,
```

src/sec_certs/dataset/protection_profile.py
```diff
@@ -417,21 +431,57 @@ def _extract_pp_keywords(self):
...
+            for dgst, record in matched.items():
+                self.certs[dgst].heuristics.scheme_data = record.to_enrichment_dict()
+
+            matched_ids = {id(v) for v in matched.values()}
+            added = 0
+            for entry in entries:
+                if id(entry) not in matched_ids:
+                    pp = ProtectionProfile.from_scheme_record(entry)
+                    if pp.dgst not in self.certs:
+                        pp.heuristics.scheme_data = entry.to_enrichment_dict()
```
jborsky: I'd move setting up the scheme_data directly to `ProtectionProfile.from_scheme_record`.

src/sec_certs/model/pp_matching.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+import re
+from typing import TYPE_CHECKING
+
+from sec_certs.model.matching import AbstractMatcher
+
+if TYPE_CHECKING:
```
jborsky: These are not heavy as well, so there's also no need to guard them.

src/sec_certs/sample/pp_scheme.py
```diff
@@ -0,0 +1,267 @@
...
+
+from sec_certs.constants import REQUEST_TIMEOUT
+from sec_certs.serialization.json import ComplexSerializableType
+
+logger = logging.getLogger(__name__)
+
+# NIAP API endpoints
+_NIAP_BASE_URL = "https://www.niap-ccevs.org"
+_NIAP_PP_API_URL = _NIAP_BASE_URL + "/api/protection-profile/public_pps_all/"
+_NIAP_PP_DETAIL_URL = _NIAP_BASE_URL + "/protectionprofiles/{}"
```
jborsky: This constant seems unused.

src/sec_certs/sample/pp_scheme.py
```diff
@@ -0,0 +1,267 @@
...
+from typing import Any, Literal, Protocol
+
+import requests
+
+from sec_certs.constants import REQUEST_TIMEOUT
+from sec_certs.serialization.json import ComplexSerializableType
+
+logger = logging.getLogger(__name__)
+
+# NIAP API endpoints
```
jborsky: I'd consider encapsulating those constants in the `NIAPScraper`.

src/sec_certs/sample/protection_profile.py
```diff
@@ -31,7 +32,15 @@ class ProtectionProfile(
 ):
     @dataclass
     class Heuristics(BaseHeuristics, ComplexSerializableType):
-        pass
+        scheme_data: dict[str, Any] | None = field(default=None)
+
+        @property
+        def serialized_attributes(self) -> list[str]:
+            # Persist scheme_data only when populated; PPs without scheme enrichment omit the key.
```
jborsky: I'd serialize it anyway; it would just be null if there's no scheme_data.

The reason I'd opt for this is that if a serialization schema were needed https://github.com/crocs-muni/sec-certs/tree/main/src/sec_certs/serialization/schemas. It would then be consistent with CC, where such fields are always present and just null when empty.

src/sec_certs/sample/protection_profile.py
```diff
@@ -159,6 +170,9 @@ def _html_row_get_date(cell: Tag) -> date | None:
 
         @staticmethod
         def _html_row_get_name(cell: Tag) -> str:
+            p_tag = cell.find("p")
+            if p_tag:
+                return p_tag.get_text().strip()
```
jborsky: Reason for this?

src/sec_certs/sample/protection_profile.py
```diff
@@ -102,6 +111,8 @@ def _from_html_row_classic_pp(
             if not sanitization.sanitize_cc_link(pp_link):
                 raise ValueError(f"pp_link for PP {pp_name} is empty, cannot create PP record")
 
+            is_collaborative = cells[0].find("p") is not None
```
jborsky: There are two functions, this one and `_from_html_row_collaborative`. These are meant to differentiate parsing for classic and collaborative PPs, so deciding `is_collaborative` inside the classic one is wrong.
 
In `from_html_row` there's branching on whether it's collaborative to call the specific function, and the `is_collaborative `flag is decided in: https://github.com/crocs-muni/sec-certs/blob/0699475cb2dc5389792783f230ebe8a6e102eb22/src/sec_certs/dataset/protection_profile.py#L233

So this function should only get non-collaborative rows; the original False was correct. Their structures are probably different, which is why there are these two functions; it feels wrong to set `is_collaborative` to True but process it as non-collaborative.

If you've found that the active/archived pages contain collaborative PPs (structure changed or whatever) that the filename check misses, fix it at the dispatch level so those rows go through the correct function.

jborsky (changes requested):
Great start, thank you 👍

I put some comments. But some probably won't be relevant anymore as the code will change based on the discussion below.

The wiring needs to be changed. Since the schemes not only enrich the metadata of existing PPs in the dataset (as in the CC schemes) but also insert new ones, the rest of the pipeline is not executed for those new ones (download, extraction, ...), and the resulting dataset has the new ones but just with metadata. So the merging step needs to be done sooner in the pipeline.

I would propose the following:

1) Move the scraping of the schemes to the `get_certs_from_web` step. First, fetch the PPs from the CC portal and then from the schemes. Perform the merge step to save additional metadata for those already present in the CC portal and insert new ones that are missing there. However, it should be okay if some of the schemes fail, since the schemes are changing, and it's common for scraping to break. With that, the dataset will be populated for other steps that remain unchanged. 

2) I'd introduce a new attribute in the PP sample named something like `other|scheme?_metadata` that would consist of the optional scheme's metadata (similar to what the EUCC sample has saved additional metadata), instead of putting it in heuristics. (Or it could be left there to semantically indicate that it's matched heuristically. I am not convinced right now what's better)

3) Consider whether we need even to serialize and store this scheme dataset, or if it will be discarded after the merge step. I don't think we need to serialize and store the scheme dataset anymore; nothing downstream consumes the raw records, and a persisted pp_scheme_data.json would just become a largely redundant metadata storage. The only reason to keep it would be if we wanted to run the matching without rescraping. If not, I would keep it as just a temp in-memory structure.

src/sec_certs/sample/protection_profile.py
m-mikulica: You were right that the classic parser shouldn't be deciding `is_collaborative`
itself, so I moved that decision up to the dispatch (`from_html_row`) and pass it
in as a parameter now.

src/sec_certs/sample/protection_profile.py
m-mikulica: the archived CC portal page is confusing because it has some
collaborative PPs mixed in with the normal archived ones, and those rows put the
name in a `<p>` instead of `<a>`. The `is_collaborative` flag comes from the file
name, so it's the same for the whole file and can't catch them. I also can't send them
to `_from_html_row_collaborative` , because that one is for the 5-column
collaborative page and these rows have 7 columns (archived layout) because it would
fail on the column check. So they go through the normal archived parser, I just
set the flag right. That's why the `<p>` check stays (now in a small helper
`_html_row_is_collaborative`), and `_html_row_get_name` still reads from `<p>`
when it's there.

m-mikulica:
Reply for your 3 main suggestions:
1. Scraping + merge now happens in `get_certs_from_web` 
2. The scraped scheme data is no longer saved to disk. The scrape now runs in-memory and is only used for the merge.
3. The enrichment now lives in a dedicated `scheme_metadata` attribute instead of inside `heuristics`.
I added a `scheme_metadata` field on `ProtectionProfile`, it's always
serialized (null when there's nothing), and it's set in `from_scheme_record`.
The old `heuristics.scheme_data` field is gone.

Smaller changes:
- Removed the unnecessary `TYPE_CHECKING` guards in `pp_scheme.py` and `pp_matching.py`.
- Fixed the broken `PPSchemeDataset` docstring and removed the unused `__getitem__`
- Removed the unused `_NIAP_PP_DETAIL_URL` constant.
- Moved the NIAP constants inside the `NIAPScraper` class as class attributes.
- (`aux_handlers` already had `| None` in its type, so nothing to change there.)
- I also made it log a clear message when the scheme enrichment is skipped
   or fails, so it doesn't disappear silently.

Tests:
I tested the whole pipeline against the live data: build + NIAP scrape + merge gives
+-600 PPs, the new US PPs get merged during the build step. Every PP now has a `scheme_metadata` key (filled for
+-270 of them, null others), and the offline tests also passed.

m-mikulica:
Until now we thought collaborative Protection Profiles only live on the dedicated collaborative page. But the regular active/archived category tables also contain some collaborative PPs. Those rows have a slightly different layout — the name sits in a `<p>` tag and the link is with label "Protection Profile" — so the old parser read the name wrong and flagged them as non-collaborative.

Fix:
The classic-row parser now detects this case - if the first cell has a `<p>`  tag, it's a collaborative PP, so we reuse the existing collaborative helpers to read the name and link. The rest of the row (version, dates, scheme…) is parsed as before.

Changes:
The classic parser decides every row whether a PP is collaborative, instead of relying only on which page it came from.
Name/link for these  rows are read with the existing collaborative getters.
Renamed the page-level flag `is_collaborative` to `from_collaborative_page` so it clearly means "this row came from the collaborative page", not "this PP is collaborative".
Added a test covering both collaborative row and a normal archived row.

src/sec_certs/dataset/protection_profile.py
```diff
@@ -188,6 +190,14 @@ def get_certs_from_web(
         self.certs = self._get_all_certs_from_html(get_active, get_archived, get_collaborative)
         logger.info(f"The resulting dataset has {len(self)} certificates.")
 
+        if to_download:
```
jborsky: If you want to guard it with some flag, which I think is a good idea, I would introduce a new flag specifically for schemes. 

With that, you can choose to download the CC portal, but not the schemes.

src/sec_certs/sample/protection_profile.py
```diff
@@ -203,12 +217,15 @@ def __init__(
...
         heuristics: Heuristics | None = None,
         state: InternalState | None = None,
+        scheme_metadata: dict[str, Any] | None = None,
     ):
         super().__init__()
         self.web_data: ProtectionProfile.WebData = web_data
         self.pdf_data: ProtectionProfile.PdfData = pdf_data if pdf_data else ProtectionProfile.PdfData()
         self.heuristics: ProtectionProfile.Heuristics = heuristics if heuristics else ProtectionProfile.Heuristics()
         self.state: ProtectionProfile.InternalState = state if state else ProtectionProfile.InternalState()
+        # Enrichment from national schemes - always serialized (null when not there)
```
jborsky: This comment is useless

src/sec_certs/sample/protection_profile.py
```diff
@@ -23,6 +23,7 @@
 
 if TYPE_CHECKING:
     from sec_certs.converter import PDFConverter
+    from sec_certs.sample.pp_scheme import PPSchemeRecord
```
jborsky: Why is this guarded?

src/sec_certs/dataset/pp_scheme.py
```diff
@@ -0,0 +1,56 @@
...
+from pathlib import Path
+
+from sec_certs.dataset.json_path_dataset import JSONPathDataset
+from sec_certs.sample.pp_scheme import PP_SCHEME_SCRAPERS, PPSchemeRecord, PPScraper
+from sec_certs.serialization.json import ComplexSerializableType
+
+logger = logging.getLogger(__name__)
+
+
+class PPSchemeDataset(JSONPathDataset, ComplexSerializableType):
```
jborsky: If we don't serialize the dataset, do you need to inherit from these classes?

src/sec_certs/sample/pp_scheme.py
```diff
@@ -0,0 +1,256 @@
...
+import requests
+
+from sec_certs.constants import REQUEST_TIMEOUT
+from sec_certs.serialization.json import ComplexSerializableType
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class PPSchemeRecord(ComplexSerializableType):
```
jborsky: Here, you probably don't need to inherit this either.

src/sec_certs/sample/pp_scheme.py
```diff
@@ -0,0 +1,256 @@
...
+        logger.info("Fetching Protection Profiles from NIAP API: %s", NIAPScraper._NIAP_PP_API_URL)
+        resp = requests.get(NIAPScraper._NIAP_PP_API_URL, timeout=REQUEST_TIMEOUT)
+        resp.raise_for_status()
+        data = resp.json()
+        logger.info("Fetched %d Protection Profiles from NIAP.", len(data))
+        return data
+
+    @staticmethod
+    def _fetch_niap_pp_files(pp_id: int) -> list[dict[str, Any]]:
+        resp = requests.get(
```
jborsky, line 112: Could you reuse some existing helpers?

E.g., this one?

https://github.com/crocs-muni/sec-certs/blob/57f072cd08d2d290a667844b069ecbe38d4b0d4d/src/sec_certs/utils/helpers.py#L99

Also, for every PP, there are two serial request calls, each with some timeout. It would be nice to make it parallel, as the repo already has this for these cases:

https://github.com/crocs-muni/sec-certs/blob/57f072cd08d2d290a667844b069ecbe38d4b0d4d/src/sec_certs/utils/parallel_processing.py#L73

src/sec_certs/sample/pp_scheme.py
```diff
@@ -0,0 +1,256 @@
...
+            return None
+        try:
+            return datetime.fromisoformat(value.replace("Z", "+00:00")).date()
+        except (ValueError, AttributeError):
+            return None
+
+    @staticmethod
+    def _niap_status_process(status: str, sunset: date | None) -> Literal["active", "archived"]:
+        # The list endpoint's status is almost always Publishing while NIAP treats a PP as archived
+        if status.lower() == "archived" or (sunset is not None and sunset < date.today()):
```
jborsky: This means a PP whose sunset date is exactly today is still classified active, and only flips to archived the following day. Is that the intended semantics? 

If it's expires as of this date, this should be <=

m-mikulica:
- Added a separate `get_schemes` flag on `get_certs_from_web` 
- Fixed sunset check: `<` → `<=` 
- Removed the unnecessary `TYPE_CHECKING` guard
- `PPSchemeDataset`/`PPSchemeRecord` no longer inherit the serialization classes 
- Parallelized the NIAP  requests with the existing `process_parallel` helper - full scrape went from 4min to 17s

jborsky (approved):

PR #666, opened 2026-06-28
fix(dashboard): show a save confirmation toast instead of "already open"

### LINKED ISSUE #568: 1.2. Improve dashboards
**Description**
Implementation of dashboards by David V.. This should be merged, tested and improved. After it seems fine for production, the “dashboard” role should be given to users by default to enable it in production.

There are [PR #540](https://github.com/crocs-muni/sec-certs/pull/540) and [PR #539](https://github.com/crocs-muni/sec-certs/pull/539) that finalize the implementation.

**Additional context**
CCAT project

xvalec01:
Addresses **point 2** of the UI review in crocs-muni/sec-certs#568.

## Problem
After clicking **Save Dashboard**, the user saw **"This dashboard is already open."** instead of a save confirmation (see review screenshot).

**Root cause** — cross-callback chaining. The toast was written *only* by the dashboard-selection callback; the save callback never emitted one. Saving a **new** dashboard set the selector value to the saved id, which re-fired the selection callback into its `dashboard_id == current` "already open" branch. Saving an **existing** dashboard returned `no_update`, so it was **silent**.

## Fix
All changes in `callbacks/dashboard.py`:
- **Save callback now owns the toast** (3 outputs via `allow_duplicate`):
  - success → **"Dashboard '<name>' saved."** for both new and existing dashboards
  - previously-silent failures (no permission / dashboard missing / not signed in) → **danger** toast
- **Neutralized the "already open" branch** in the selection callback so the post-save re-trigger emits no toast (and can't clobber the success toast). Re-selecting the already-current item doesn't fire a change, so the branch had no other real-world trigger.

### DIFF
--- a/sec_certs_page/dashboard/callbacks/dashboard.py
+++ b/sec_certs_page/dashboard/callbacks/dashboard.py
@@ -201,7 +201,10 @@ def _register_dashboard_selection(
                 "toast_icon": "info",
             }
 
-        # Check if the selected dashboard is already open
+        # The selected dashboard is already open. This branch is primarily hit as a
+        # re-trigger after saving (save sets selector_value to the saved id), so it
+        # must not emit a toast - doing so would clobber the save-success toast with
+        # a misleading "already open" message. Leave the view and toast untouched.
         if current_dashboard_id and str(dashboard_id) == str(current_dashboard_id):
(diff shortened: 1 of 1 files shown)


MaryBak:
LGTM

jborsky (approved):

Issue #579, opened 2026-03-17
2.2. Improve the usability of the references (graph and list)


MaryBak:
**Description**
* Add the legend for the reference graph, make the highlight of the reviewed certificate more distinguishable.
* Clean up the visual clutter and "jittery" movement in the graph 
* Redesigning the reference lists to clearly distinguish between incoming and outgoing document links and explain what those mean

**Details for underlying issues and ideas for implementation:**
* Add the legend for the reference graph: the edge direction meaning, dots, colors, icons [R29, 30, 32, 33]
* Make the highlight of the reviewed certificate more distinguishable (with usage of a separate color or shade) in the reference graph [R31] 
* The reference graph is very cluttered, the force based layout of nodes can move in an unsettling way (until it settles) and the arrows of the edges are mostly not visible (because they are drawn under the icons of the certificates). Similarly, the listing of incoming and outgoing references of a certificate is confusing. They are displayed next to the document that they are contained in, even if they are incoming references, meaning that it is the document of the other certificate that they are present in.
* When you open the references using CC->References, the browser freezes. Testing setup: processor AMD Ryzen 7 PRO 8840U w/ Radeon 780M Graphics, RAM 32 GB; OS: windows 11; browser: Brave

**Additional context**
CCAT project

PR #668, opened 2026-06-30
1.1. Rework search UX

### LINKED ISSUE #567: 1.1. Rework search UX
**Description**
Fixing the search to work like a modern e-shop, adding sidebar filters for labs, manufacturers, and schemes so you don't have to rely on tricky text queries. This update unifies the search syntax, fixes "verbatim" quotes to actually work, and adds clickable columns for easy sorting. We’ll also add a "no results" help guide.

* **Search syntax** is not clear to users and is not the same for name and fulltext search.
* “**Verbatim search**” in quotes does not work fully verbatim (i.e., there is still stemming and other transformations applied to the text and query).
* There is no way to revert the **sort direction** (ascending/descending).
* In the search results, the properties of the shown certificates (ID, name, cert. date, etc.) are fixed and may not be what the user wants. The CC and PP pages are for example missing the scheme of the certificate.
* The limitations of the search are not clear to the users. For example, (i) the fact that we source everything from CC portal and that some schemes do not put everything on CC portal, (ii) the limitations of the database/search results (starting from which date the database has been populated and which certificates are excluded) [R2]
* Even though we extract a lot of structured data from the certificates, we do not provide a way to search or filter certificates using the fields. For example, one would like to search by evaluation lab, or see all certificates that we match the “SHA” cryptographic primitive in (without using full-text search).
* Full-text search of a single certificate (or a subset of certificates) is not possible.
* The CVE and CPE search fields ignore an enter key press and are not “real” search functions; they merely navigate to the given text string, thus no substring search is possible.
* Then searching for a non-existent certificate in CC, the result set is empty, but the search results display an empty graph and a result table header without any items.
* Move sorting from the advanced options area to the area between search and the results. [R20, 22]
* Sort the search results by default from the newest at top to the oldest at the bottom [R21]
* Add the information about the “display network” function. [R23]
* Unify the structure of search results in FIPS and CC searches [R24]
* Add search by new specific fields: the national scheme, manufacturer, and certification lab [R3]
* **[UPD 23.03.2026]** By default, all the available certificates are presented in the results (before interacting with the search bar), but it's not possible to neither to filter this whole dataset nor search it using the search bar (unless using the website tabs to navigate to the initial search state).
* **[UPD 25.03.2026]** Adding more details about the product makes the search wider, not more specific, and the results are hard to parse. 
     * E.g., if you search for general model "Xerox" in FIPS, you get 0 results, but if you specify specific product details (e.g., "Xerox WorkCentre Pro 232"), you get 69 results, but most of them are irrelevant. The issue is caused by the fact that each new word added to the search query adds new matches.
     * One way to decrease cognitive load on the user when they parse the results (and are confused with the results which seem to be completely irrelevant) is to highlight the found matches in the result table. 

Image

**Additional issues should be considered** 
The search should be overhauled keeping in mind:
* Ideally, some pre-design exploration of search and filtering of similar complexity of items should be done. Perhaps parts of the functionality (the keyword and other filtering) could be similar to what e-shops do (e.g. [Alza](https://www.alza.cz/notebooky/18842920.htm) in the left panel).
* Allow for changing sort direction.
* Allow for selecting sort column by clicking on it (the usual UI).
* Perhaps allow for selecting which columns to display.
* Provide multiple search syntax examples, perhaps have them cycle through after a few seconds of being shown.
* Do not show the table/graph for empty result set.
* Add a generic comment after the search results roughly going like "Didn't find what you were looking for? See <here> for some tips and tricks." And that link would be to some FAQ page or a modal which would explain some common issues with the search and also provide info about the data source.
* Turn CVE and CPE into real search.
* Perhaps add a PDF viewer into the page and make full-text search results link to highlights directly into the PDF on the page.
Consider using just the full-text search tool Whoosh to also do the name search (it already can do this). This would unify the syntax and simplify the UI.
Consider changes to the tokenizer and query evaluator for Whoosh full-text search to improve it.

**Additional context**
CCAT project

jborsky:
This PR addresses https://github.com/crocs-muni/sec-certs/issues/567:

It's a large rewrite of the search. It migrates to the new search engine, replaces the old search code, templates, and indexing, and improves the search UI.

Main changes:
- Migrate search from Whoosh + MongoDB to just Tantivy, and completely refactor/unify the search code
- Add more filtering options, including text fields, a keyword tree for filtering on extracted keywords, range sliders, and multi-select facets
- You can use a full-text search combined with filters, so you can search within any filtered subset
- Add CVE/CPE search
- Add sortable columns and a column picker
- Swap the results in place instead of a full page reload
- Make the UI more compact and nicer in general

Things that still need to be discussed and potentially done in follow-up PRs:
- Improve the query syntax guide with examples and a "no results" FAQ.
- The search currently uses the default Tantivy tokenizer; implement, tune, and test a custom one (the old one that was used can be used as a basis) 
- Considered merging the name and full-text search altogether
- Add back an option to display the references graph for a subset of certificates (should be done with https://github.com/crocs-muni/sec-certs/issues/579)
- Timeline was removed. Do we want it back? If so, it will require review and improvement, as the axes looked weird in some cases.


jborsky:
@J08nY do you want to take a look?

PR #673, opened 2026-07-03
Fix: EUCC cert_id bug

m-mikulica:
FR/NL certificates published under EUCC on commoncriteriaportal.org get
wrong `heuristics.cert_id`: many different French EUCC certs all have the same `ANSSI-CC-3090/20` cert_id.
It also corrupted relations among certs.

1. Old FR/NL filename rules are wrong on EUCC filenames — the FR rule
   matches `3090-20` inside `EUCC-3090-2025_10_04_Rapport.pdf`
 NL rule matches the `CC-` substring inside `EUCC-`.
2. The correct EUCC ID never gets through because the existing `EUCC-…$` rule
   ends with "$" and only knows the long form.

Fix 
- `rules.yaml`: new EUCC rules for FR (3090) and NL (3110) covering short and
  long forms, with different separators and spaced text 
- `cc_certificate_id.py`: builds canonical `EUCC-<CB>-<year>-<seq>`

Tested:
- New unit tests
- On a production dump all 64 affected FR+NL certs got correct IDs
-PDF extraction: keyword extraction now finds the EUCC id in text
Works also on three certs whose filenames carry no extractable id


src/sec_certs/sample/cc_certificate_id.py
```diff
@@ -19,7 +19,24 @@ def _parse_year(year: str | None) -> int | None:
...
+    return None
+
+
 def FR(meta) -> str:
+    # EUCC formsEUCC-3090-2025-10-04 → EUCC-3090-2025-4
+    if eucc := _eucc_id("3090", meta):
+        return eucc
+    # EUCC-ANSSI form
+    if meta.get("month") and meta.get("day"):
+        return f"EUCC-ANSSI-{meta['year']}-{meta['month']}-{meta['day']}"
```
J08nY: This also changes all past ANSSI ids into EUCC ids. I do not think that is correct. They are not EUCC certificates... Peraps we need to introduce another CertificateId class for EUCC.

jborsky:
Thanks for your effort on this. 

Some of the changes contain hidden decisions I'd like to discuss first. To break this issue down and provide a context:

We have some EUCC certificates taken from the ENISA page and we also see a bunch of EUCC certificates published on the CC portal through the individual schemes. In FR and NL schemes, this caused issues with assigning the correct certificate ID as the rules weren't updated yet to handle these new IDs. That's the bug we are solving.

Now, speaking of the assignment of the ID itself, when looking into this in more depth, there are several things we need to decide how to handle, and based on those decisions, we add/edit the extraction rules and the assignment.

In general, the EUCC ID format should be:

`EUCC-<body>-<year>-<seq>-<revision>`

Now let's look at how each scheme actually produces the IDs:

#### ENISA

Example:

`EUCC-3090-2025-0000000006-00000`
(https://certification.enisa.europa.eu/certificates/eucc-3090-2025-0000000006-00000_en)

It has a fixed 10-digit zero-padded sequence number and explicit revision.

#### FR (body 3090)

Example: 

`EUCC-3090-2026-32`
(https://www.commoncriteriaportal.org/nfs/ccpfiles/files/epfiles/EUCC-3090-2026-32-rapport.pdf)
(Not on ENISA yet)

For certs in 2026 it follows the general format, just without zero-padding and without an explicit revision number.

2025 was a wild time, they have a variants with month included:

`EUCC-3090-2025-11-08`
(https://www.commoncriteriaportal.org/nfs/ccpfiles/files/epfiles/EUCC-3090-2025-11-08-rapport.pdf)
(https://certification.enisa.europa.eu/certificates/eucc-3090-2025-0000000008-00000_en)

or with the acronym used instead of the NANDO number:

`EUCC-ANSSI-2025-03-02`
(https://www.commoncriteriaportal.org/nfs/ccpfiles/files/epfiles/EUCC-ANSSI-2025_03_02fr.pdf)
(https://certification.enisa.europa.eu/certificates/eucc-3090-2025-0000000001-00002_en)

#### NL (body 3110)

Example:

`EUCC-3110-2025-08-2500053-01`
(https://www.commoncriteriaportal.org/nfs/ccpfiles/files/epfiles/EUCC-3110-2025-08-2500053-01_Cert.pdf)

They include the month, and here we also see the revision number (-01). You'd expect the mapping into the ENISA long format to just be "drop month, zero-pad the seq, zero-pad the revision":

`EUCC-3110-2025-0002500053-00001`
(https://certification.enisa.europa.eu/certificates/eucc-3110-2025-0002500053-00001_en)

…and for that cert it is. But then there's this case:

`EUCC-3110-2026-01-2500076-01`

Which ENISA mapped to

`EUCC-3110-2026-2500076-01-00000 `
(https://certification.enisa.europa.eu/certificates/eucc-3110-2026-2500076-01-00000_en)

So here they dropped the month, didn't zero-pad the seq, kept the scheme's revision -01, and then appended their own -00000 on top (??)

#### DE (body 3087)

They publish EUCC certificates as well, but they also keep their classic ID, so we are not broken yet. I found 4 of them on sec-certs. 

Example:

https://sec-certs.org/cc/db6e00b041b6c717/

From the report front page:

`EUCC-3087-2026-04-0004`

(month + 4-digit seq, no revision):

Matching ENISA one (month dropped, 10-digit padded seq, appended revision):
https://certification.enisa.europa.eu/certificates/eucc-3087-2026-0000000004-00000_en

Ughh, context given.

---
Now, some things to discuss:

This PR adds FR/NL parsing rules for the formats above (haven't verified them for correctness yet) and then canonicalizes the IDs into:

`EUCC-<CB>-<year>-<seq>`

Before we lock that in, a few things are worth discussing, because the decision is now baked in the change.

**1. Do we actually need canonicalization for ID assignment?**

From a quick look, within a single certificate the EUCC id string is essentially the same across that cert's own files (report / ST / cert). Maybe only when filenamse has `_` instead of `-` as seperator, but simple normalization would handle it. So, for assignment canonicalization isn't needed.

However, I understand that it can be useful for having a consistent format across different bodies, so...

**2. What should the canonical format be?**

Short format, long padded format?

Should we include the revision?  Why not? We could always track the revisions anyway. But do we want to have the same assigned ID for them or a distinct one?

**3. Do we canonicalize the ENISA long forms too, and do we want to link CC ↔ EUCC?**

Do we also want to use the canonicalization for our EUCC section? Or we will keep the ones there to be consistent with ENISA? 

We could also link the matching certificates between the CC and EUCC sections on our page (this could also be done by keeping the long formats there, of course). So, in general, would we like to link them?

@tkachyna thoughts? (as you are the EUCC guru) @J08nY ? @MaryBak ?

src/sec_certs/sample/cc_certificate_id.py
m-mikulica: I understand, that its confusing. This dispatch distinguishes 3 forms of French cert ids.
1.EUCC form - the new official form
2.EUCC-ANSSI form - transitional form, that was after old ANNSI and before new EUCC form
3.ANNSI - old classic form
The old classic ANSSI id is handled by the unchanged code path (lines 40-53), and
this fix does not change its behaviour.What the fix adds is handling for forms 1 and 2, which previously weren't handled at all.

J08nY:
> **1. Do we actually need canonicalization for ID assignment?**

Yes. The original reason for canonicalization is that various parties/documents may refer to the same certificate using slightly different IDs. For example when ANSSI changed the naming from DCSSI-XX/YYYY to ANSSI-CC-XXXX-YY or something like that. Some documents along the transitional period used the old, some used the new. Thus, to have the references between the certificates to be correct, you need to canonicalize the ids somehow. Or at least, you need to create some equivalence classes on the set of IDs, such that the above two ways of referring to the same certificate are equivalent. We decided to just canonicalize the IDs into some middle-ground form, also fixing some errors along the way.


> **2. What should the canonical format be?**
> 
> Short format, long padded format?
> 
> Should we include the revision? Why not? We could always track the revisions anyway. But do we want to have the same assigned ID for them or a distinct one?

No idea. Depends on what the revisions mean. The certificate should just have one ID and no two certificates should have the same ID. If you strip the revision, can some of the two conditions just mentioned be broken?


> **3. Do we canonicalize the ENISA long forms too, and do we want to link CC ↔ EUCC?**
> 
> Do we also want to use the canonicalization for our EUCC section? Or we will keep the ones there to be consistent with ENISA?
> 
> We could also link the matching certificates between the CC and EUCC sections on our page (this could also be done by keeping the long formats there, of course). So, in general, would we like to link them?

The answer to "do we want more information" is almost always yes. The question is how. I do not know the relation between CC and EUCC and what it means for us (unfortunately). And the relation between scheme websites, ENISA and the CC portal is I guess even less formally defined and messy. In that case, my thinking is, capture all the information, put it in some sensible canonical format (that does not lose information) and leave it there.

jborsky:
> The answer to "do we want more information" is almost always yes. The question is how. I do not know the relation between CC and EUCC and what it means for us (unfortunately). And the relation between scheme websites, ENISA and the CC portal is I guess even less formally defined and messy. In that case, my thinking is, capture all the information, put it in some sensible canonical format (that does not lose information) and leave it there.

To link CC certificates to EUCC certificates is probably a much more difficult question, and I don't know how to either. 

What I meant is the relation for the same EUCC certificates between scheme websites (EUCC certificates published on the CC portal, so in our CC section, to certificates in our EUCC section scraped from ENISA). How is "simple" in this case. They are the same certificates, and if we canonicalize the IDs in both sections, they will have the same ID (should have at least for most of them). Or if we decide not to show the canonicalization in the EUCC section to be consistent with ENISA, we can still canonicalize them during linking.

J08nY:
Yes, I meant the same relation. When the same certificate appears on multiple sources that we scrape and hence creates multiple entries for us. We should recognize and link the two (certificate from CC portal and certificate from ENISA/scheme website).

m-mikulica:
1. Do we actually need canonicalization for ID assignment?
I did it mainly because I considered it important for the certificate relations to other certs. If cert A would cite cert B and both of them would have different style of id (EUCC-3090-2025-4 vs EUCC-3090-2025-10-04) the matching may not work properly.
2. What should the canonical format be?
I decided to use the shortest format, because its most relaible. We can always make it from the longer ones. 
Also I throwed away the revision number, because some of the certs simply does not have it(e.g all FR certs). 

I think that, the revision  does not have to be lost. The NL regexes already capture it
as a named group (eucc_rev), its just discarded during canonicalization. As a
future improvement we could persist it into a heuristics field. Since the DB primary key
is the digest, two revisions would be kept
as two entries anyway and the cert page could then list the sibling revisions
by looking up other certs sharing the same cert_id.

tkachyna:
> 1. Do we actually need canonicalization for ID assignment?

I agree with what Jano wrote. I just looked back at the ENISA website from the last time they unified the IDs. The file names (cert, cert report, and ST) actually remained the same, and what's even worse is that on the CC portal the certificates in the table still use the original IDs. I don't have enough insight into how the references work internally or whether there is any tolerance for differences in IDs, but I can easily imagine this causing quite a mess, especially if ENISA decides to get creative with the IDs again in the future. That said, I'm now about 99% confident that they will remain stable going forward. I also think that canonicalizing the IDs will be easier than implementing heuristics to match these different ID formats.

If I were to suggest a convention we should follow, I'd keep the `EUCC <body> <year> <seq>` part, but I'd remove the leading zeros from the sequence number for readability. The remaining question is what to do with revision numbers. As long as we don't keep older versions of certificates (although I remember there was some discussion during the last meeting that we could start doing that), including the revision number seems unnecessary to me.

At the same time, for clarity, I'd still preserve the original certificate ID somewhere, either in smaller text or in a tooltip.

> 2. What should the canonical format be?

Answered in the previous answer.

> 3. Do we canonicalize the ENISA long forms too, and do we want to link CC ↔ EUCC? Do we also want to use the canonicalization for our EUCC section? Or we will keep the ones there to be consistent with ENISA?

I am for it, for the linking. I'd try to keep things consistent by applying the same approach in the EUCC section. I also think it's important to have a single source of truth for all EUCC certificates. For certificates that are available on both the cc portal and ENISA, I'd treat ENISA as the authoritative source in terms of IDs (and as I mentioned earlier, I'd make a small adjustment to these ENISA IDs by removing the leading zeros from the sequence number and, for now, probably omitting the revision number as well),

jborsky:
Thanks to all for your thoughts.

- Let's keep the proposed canonical format, but also apply it in the EUCC processing and save the original form separately somewhere for now. @m-mikulica can you do this change? Once it's done, I'll check the rules for correctness.
 
- Once we have the correct IDs, we can link them across sections on our page. The datasets now are independent, so we can't treat either one as an authoritative source for IDs. We can only link entries that happen to share an ID. But let's keep this out of the scope of this PR. What to do with this in the future is up for discussion, as I also don't know the relationship between CC and EUCC or what that implies for how we should process and present the data.

m-mikulica:
Applied the canonical format to the EUCC processing

Changes:
- EUCC certificates now receive their canonical id in `heuristics.cert_id` (the same short form used in the CC section), while the original ENISA id is saved in the top-level `cert_id`
- Previously `heuristics.cert_id` was always `None` in the EUCC section, so now its a good starting point for matching between certs
- New helper `canonicalize_eucc()` ,which is scheme-independent (rather than reusing the per-scheme FR()/NL() functions) because the certification body is already encoded in the id, so no scheme lookup is needed.

Results:
- 38 canonicalized, 37 unique
- 20 now share a canonical id with a CC-section certificate

Known issue:
https://certification.enisa.europa.eu/certificates/eucc-3090-2025-0000000001-00002_en
https://certification.enisa.europa.eu/certificates/eucc-3090-2025-0000000001-00001_en
ENISA assigned them the same sequence number, even though they are different products. They used the revision number to differ them, so I think this is not our fault and it's not a thing we could do much about.
So now the two remain separate entries (distinct `dgst`) and only share the `heuristics.cert_id`

jborsky:
@m-mikulica I've pushed some commits to make it more general and nicer. Can you look at it and test it ?

Also, two tests now fail because I changed the rule a little bit and no longer expect `_` as a separator in the `cc_cert_id` rules. I kept it just for filenames rules. 

Could you check if there are any instances in the documents where an underscore is used? If you find any, please add it as a possible separator in the rules. If not, please update the tests.

m-mikulica:
Little changes:

1. I checked the PDFs for the `_`, but it seems that it only appears in the filenames, so I just changed the tests.

2. Scheme detection fix for NL
ENISA sometimes writes NL ids with two revision blocks at the end(https://certification.enisa.europa.eu/certificates/eucc-3110-2026-2500075-01-00000_en)
The new NL rule only expects one rev block, so the whole match failed. So I slightly changed the rule and now the 5 broken certs have their correct scheme. This is also added to the tests.

3. CCSchemeMatcher stopped getting cert ids from EUCC filenames
It seems adding the new `_eucc_id(meta)` dispatch to the matching mechanics was just forgotten, so now it's there.
I tested it on my local database and everything works well I think.

PR #686, opened 2026-07-29
Update EUCC regexes

tkachyna:
- certs with new regexes have dropped, needed to update them a little bit (DE, SE, ES). Mainly add a new nando ids and milder restrictions for some segments


jborsky (approved):

J08nY:
image

PR #689, opened 2026-08-15
2.7. Improve accuracy of scraping NSCIB Common Criteria report frontpages

### LINKED ISSUE #584: 2.7. Improve the accuracy of heuristic data extracted from the CC front page
**Description**
Reimplement frontpage parsing for CC. This will improve some of the heuristics we report, such as the evaluation lab or certificate ID, since the front page is very data-rich.

There is significant technical debt in the way we parse CC certification report frontpages. Issue [#533](https://github.com/crocs-muni/sec-certs/issues/533) describes this. This should all be refactored and enhanced. Doing so will nicely improve some heuristics we report, like the evaluation lab or certificate ID, as the frontpage is very data-rich.

**Additional context**
CCAT project

tkachyna:
**Summary:**
- Introduces an abstract Frontpage parser class
- Introduces concrete implementation for NSCIB CC report frontpages
- Other countries will be added in upcoming PRs
- Adds tests

**Detailed description:**
Introduces sec_certs/utils/frontpage.py with a FrontpageParser ABC and a concrete NSCIBFrontpageParser. The design normalizes both converter outputs into the same shape before any parsing happens:                                   
                                                                                                                                                                                                                                         
  ```mermaid                                                                                                                                                                                                                             
  flowchart LR                                                                                                                                                                                                                           
      A[pdftotext .txt] --> B[_page_one_lines]                                                                                                                                                                                           
      C[Docling .json] --> D[_page_one_blocks]                                                                                                                                                                                           
      B --> E[list of str]                                                                                                                                                                                                               
      D --> E                                                                                                                                                                                                                            
      E --> F[_clean_blocks]                                                                                                                                                                                                             
      F --> G[_blocks_after_title]                                                                                                                                                                                                       
      G --> H[_collect_fields]                                                                                                                                                                                                           
      H --> I[_values_at]                                                                                                                                                                                                                
      I --> J[fields dict]                                                                                                                                                                                                               
  ```                                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                       
  - _page_one_lines / _page_one_blocks — reduce either converter to a plain list of strings                                                                                                                                              
  - _clean_blocks — strips markdown/HTML noise and empty blocks                                                                                                                                                                          
  - _blocks_after_title — drops the letterhead; None if not an NSCIB page                                                                                                                                                                
  - _collect_fields — text before the first Label: line becomes the product name                                                                                                                                                         
  - _values_at — reads the value same-line (pdftotext) or next-block (Docling); trims a merged address off company names                                                                                                                 
  - _organisation_of — legal-suffix regex first, house-number cut as fallback                                                                                                                                                            
  - _recover_cert_id — whole-page regex sweep via the scheme's rules["cc_cert_id"]["NL"]          

Field extraction is driven by one table, FRONTPAGE_LABELS, mapping printed labels to tag(s).

### DIFF
--- a/src/sec_certs/utils/frontpage.py
+++ b/src/sec_certs/utils/frontpage.py
@@ -176,9 +176,8 @@ class NSCIBFrontpageParser(FrontpageParser):
 
         if constants.TAG_CERT_ID not in items_found:
             cert_id = self._recover_cert_id(" ".join(blocks))
-            if not cert_id:
-                return {}
-            items_found[constants.TAG_CERT_ID] = cert_id
+            if cert_id:
+                items_found[constants.TAG_CERT_ID] = cert_id
 
         if product_name_blocks:
             items_found[constants.TAG_CERT_ITEM] = _clean_identity(" ".join(product_name_blocks))
--- a/src/sec_certs/utils/frontpage.py
+++ b/src/sec_certs/utils/frontpage.py
@@ -200,19 +200,28 @@ class NSCIBFrontpageParser(FrontpageParser):
         seen_label = False
 
         for idx, block in enumerate(blocks):
-            match = self._LABEL_RE.match(block)
-            if match is None:
+            labelled = self._label_of(block)
+            if labelled is None:
                 if not seen_label:
                     product_name_blocks.append(block)
                 continue
 
             seen_label = True
-            for tag, value in self._values_at(blocks, idx, match):
+            label, match = labelled
+            for tag, value in self._values_at(blocks, idx, label, match):
                 items_found.setdefault(tag, value)
 
         return items_found, product_name_blocks
 
-    def _values_at(self, blocks: list[str], idx: int, match: re.Match[str]) -> list[tuple[str, str]]:
+    def _label_of(self, block: str) -> tuple[str, re.Match[str]] | None:
+        """Return the known frontpage label of ``block`` with its match, or None if it carries none."""
+        match = self._LABEL_RE.match(block)
+        if match is None:
+            return None
+        label = match.group("label").strip().casefold()
+        return (label, match) if label in self.FRONTPAGE_LABELS else None
+
+    def _values_at(self, blocks: list[str], idx: int, label: str, match: re.Match[str]) -> list[tuple[str, str]]:
         """
         Resolve the tag/value pair(s) for the label matched at ``blocks[idx]``.
 
@@ -220,7 +229,6 @@ class NSCIBFrontpageParser(FrontpageParser):
         following block instead, sometimes with the organization's postal address merged in, which
         is trimmed off so that both converters agree.
         """
-        label = match.group("label").strip().casefold()
         value = match.group("value").strip()
         address_merged = False
 
@@ -231,7 +239,7 @@ class NSCIBFrontpageParser(FrontpageParser):
             value, address_merged = nxt, True
 
         resolved = []
-        for tag in self.FRONTPAGE_LABELS.get(label, ()):
+        for tag in self.FRONTPAGE_LABELS[label]:
             trimmed = self._organisation_of(value) if address_merged and tag in self._ORGANISATION_TAGS else value
             resolved.append((tag, _clean_identity(trimmed)))
         return resolved
--- a/src/sec_certs/utils/frontpage.py
+++ b/src/sec_certs/utils/frontpage.py
@@ -1,0 +1,266 @@
+from __future__ import annotations
+
+import html
+import json
+import logging
+import re
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import Any
+
+from sec_certs import constants
+from sec_certs.cert_rules import rules
+from sec_certs.utils.extract import load_text_file, normalize_match_string
+
+logger = logging.getLogger(__name__)
+
+# Docling labels that carry running headers/footers rather than page content.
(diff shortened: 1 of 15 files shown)


tkachyna:
Just fyi, in the upcoming PR which are going to cover rest of the country schemes, I am planning moving this abstract class to its own file and each of the scheme as well.

jborsky:
Hi, thank you. I'll probably review it tomorrow.

jborsky:
> Hi, thank you. I'll probably review it tomorrow.

Oops, that didn't happen 😄 Reviewing soon, sorry for the wait.

src/sec_certs/utils/frontpage.py
```diff
@@ -0,0 +1,266 @@
...
+    )
+
+    def parse_structured(self, document: dict[str, Any]) -> dict[str, Any]:
+        return self._parse_frontpage_blocks(self._page_one_blocks(document))
+
+    def parse_text(self, text: str, text_with_newlines: str) -> dict[str, Any]:
+        return self._parse_frontpage_blocks(self._page_one_lines(text_with_newlines))
+
+    @staticmethod
+    def _page_one_blocks(document: dict[str, Any]) -> list[str]:
```
jborsky, line 148: This may also be useful for other parsers in the future. If that is the case, you can then move this into the DocumentView as some general get text blocks function for Docling, or transfer it to the base class.

src/sec_certs/utils/frontpage.py
```diff
@@ -0,0 +1,266 @@
...
+
+class FrontpageParser(ABC):
+    """Base class for scheme-specific certification report frontpage parsers."""
+
+    certification_body: str
+
+    @abstractmethod
+    def parse_structured(self, document: dict[str, Any]) -> dict[str, Any]:
+        """
+        Extract the frontpage fields from a serialized codling document.
```
jborsky: Typo in codling

src/sec_certs/utils/frontpage.py
```diff
@@ -0,0 +1,266 @@
...
+    def parse(self, txt_path: Path, json_path: Path | None = None) -> dict[str, Any]:
+        """
+        Extract the frontpage fields, preferring the structured document when one is available.
+
+        :param txt_path: Path to the text rendering of the report.
+        :param json_path: Path to the serialized docling document, if the converter produced one.
+        :return: The extracted fields, empty if nothing could be extracted.
+        :raises ValueError: If parsing fails unexpectedly.
+        """
+        if json_path is not None and json_path.exists():
```
jborsky, line 76: For now, this branching based on json_path is ok. However, if someone were to implement different converters in the future, it would make sense to use the DocumentView here and its functions to obtain the blocks you need.

src/sec_certs/utils/frontpage.py
```diff
@@ -0,0 +1,266 @@
...
+        seen_label = False
+
+        for idx, block in enumerate(blocks):
+            match = self._LABEL_RE.match(block)
+            if match is None:
+                if not seen_label:
+                    product_name_blocks.append(block)
+                continue
+
+            seen_label = True
```
jborsky, line 212: This is set even when we don't match any tags below.

I don't know the assumptions about the format of the front pages, but this could happen if you have a product name block that would match the label regex. Wouldn't it be more robust to set seen_label only when _values_at returns some tags with values?

src/sec_certs/utils/frontpage.py
```diff
@@ -0,0 +1,266 @@
...
+
+        body = self._blocks_after_title(blocks)
+        if body is None:
+            return {}
+
+        items_found, product_name_blocks = self._collect_fields(body)
+
+        if constants.TAG_CERT_ID not in items_found:
+            cert_id = self._recover_cert_id(" ".join(blocks))
+            if not cert_id:
```
jborsky: Why do we discard everything we potentially parsed when we don't match any cert_id?

jborsky (commented):
Good job, I like it :) 

Have you tried running the extraction on our current dataset and exploring the results?

tkachyna:
So far I have tried running it isolated with randomly chosen certification reports (which we already scrape). I will try it in the next couple of days on the dataset. In the meantime I will create PRs for the next schemes.

src/sec_certs/utils/frontpage.py
tkachyna: When #691 gets merged, I will try to refactor it, as it is not in master yet.

src/sec_certs/utils/frontpage.py
tkachyna: Thanks for the info, I will make a note for myself about it.

src/sec_certs/utils/frontpage.py
tkachyna: Hmmm... good point. For some reason I thought that if the ID wasn't there, we should just discard it because it wouldn't be the cert report we want. But yeah, lets keep it.

tkachyna:
I also extended the logic for cert lab parsing, as it extracted part of the address for one specific cert I chose. I don't think we can guarantee 100% correctness since the set of lab names + addresses is really diverse, but this should at least improve it a bit.

tkachyna:
Tested in a pipeline with 25 random CC certs and EUCC certs. Example:
image

PR #690, opened 2026-08-18
Incremental updating 

### LINKED ISSUE #570: 1.4. Tweak the integrated AI-enabled PDF parser using Docling
**Description**
Tweak already integrated Docling processing and decide if we want to keep both approaches (original one and this new one)
Some issues arising from the Docling approach need to be handled:
(i) parsing FIPS IDs, which are confused for Markdown headings; (ii) the quality of Docling output is sub-par for some documents; (iii) handle the memory issues of Docling when processing large datasets.

Currently, we have two ways to convert PDF documents to text: Docling and pdftotext. Realistically, there is a third one tesserract-ocr, which is used in the pdftotext approach if its output is determined to be faulty. Furthermore, Docling (and even pdftotext) has parameters that change their behavior. Docling provides a structured output, can extract tables natively, and embedded images. However, it has a significantly increased runtime (and is a large, somewhat cumbersome dependency). We should decide on the plan going forward and answer some questions, mainly:
* Do we keep both approaches or migrate fully to Docling?
* If we keep both approaches, how do we handle changes/improvements that rely on the text being from one or the other converter?
* If we migrate fully to Docling, how do we ensure that setup and running the tool remains accessible to users?
* If we keep both, perhaps we can use one as a fallback for the other, also taking the raw tesserract-ocr into account.

In any case, some issues and opportunities arising from the Docling approach need to be handled:
* Issue [#542](https://github.com/crocs-muni/sec-certs/issues/542) where parsing FIPS ids which are confused for Markdown headings.
* Issue [#543](https://github.com/crocs-muni/sec-certs/issues/543) where quality of Docling output is sub-par for some documents.
* Handle the memory issues of Docling when processing large datasets: [PR #525 discussion](https://github.com/crocs-muni/sec-certs/pull/525#discussion_r2541524696).
* Consider dropping tabula-java and instead use Docling for table extraction in FIPS (essentially tying FIPS table extraction to having Docling installed instead of tabula-java).
* We can utilize document structure provided by Docling (see below for a more involved description).

**Additional context**
CCAT project

jborsky:
First part of https://github.com/crocs-muni/sec-certs/issues/570.

This PR introduces an option to retain the state associated with artifacts from the existing certificates in the dataset when fetching new metadata. This allows for converting and extracting data only from certificates that are either new or have changed artifacts. 

With this, Docling can be used on the server, as we will convert only a few documents during the weekly update, rather than processing the entire dataset, which could take several days.

Most of the functionality was already there (the states and hashes for artifacts, fresh flags in the pipeline steps, and so on). I just added what was missing so it can actually be used.

Summary of changes:

- Add a `carry_processing_results` option to `get_certs_from_web` across all four datasets. When set, `state` and `pdf_data` (and `web_data` for FIPS) of the certificates already held are carried over onto the freshly scraped ones, so the metadata scrape no longer discards everything that was processed before.
- Reconcile the carried over state against the files on disk. Each artifact is compared with the hash recorded when it was processed, and a missing or mismatched file invalidates the stage that produced it and everything derived from it.
- Compare the source hash on download and reset the conversion when it differs.
- Make the extraction step incremental with a `fresh` flag, mirroring the flag the other steps already had.
- Move the FIPS algorithms from `heuristics.algorithms` into `pdf_data`, and compute their union in the heuristics step. Heuristics are not carried over, so a certificate whose policy hasn't changed would never re-run the policy table extraction and would silently lose them.
- Add an `update()` orchestrator running the whole pipeline the way it is meant to be run in this mode: all artifacts are downloaded fresh, since that is the only way to learn whether a document changed, and the remaining steps then run with `fresh=False,` so they process only what is new, changed or previously failed. Heuristics are always recomputed as it depends on the whole dataset.

Follow-ups:
- Adding tests
- Wiring on page

### DIFF
--- a/src/sec_certs/dataset/cc.py
+++ b/src/sec_certs/dataset/cc.py
@@ -386,6 +386,7 @@ class CCDataset(Dataset[CCCertificate], ComplexSerializableType):
         keep_metadata: bool = True,
         get_active: bool = True,
         get_archived: bool = True,
+        carry_processing_results: bool = False,
     ) -> None:
         """
         Downloads CSV and HTML files that hold lists of certificates from common criteria website. Parses these files
@@ -395,9 +396,16 @@ class CCDataset(Dataset[CCCertificate], ComplexSerializableType):
         :param bool keep_metadata: If CSV and HTML files shall be kept on disk after download, defaults to True
         :param bool get_active: If active certificates shall be parsed, defaults to True
         :param bool get_archived: If archived certificates shall be parsed, defaults to True
+        :param bool carry_processing_results: If the dataset already holds certificates, carry their
+            already computed processing results over onto the freshly scraped certificates. So only certificates that are new,
+            changed or if files for them are missing can get reprocessed,
+            not the whole dataset. Defaults to False.
         """
         if to_download is True:
             self._download_csv_html_resources(get_active, get_archived)
+
+        old_certs = self.certs
+        self.certs = {}
 
         logger.info("Adding CSV certificates to CommonCriteria dataset.")
         csv_certs = self._get_all_certs_from_csv(get_active, get_archived)
@@ -412,6 +420,9 @@ class CCDataset(Dataset[CCCertificate], ComplexSerializableType):
 
         if not keep_metadata:
             shutil.rmtree(self.web_dir)
+
+        if carry_processing_results:
+            self._carry_processing_results(old_certs)
 
         self._set_local_paths()
         self.state.meta_sources_parsed = True
--- a/src/sec_certs/dataset/cc.py
+++ b/src/sec_certs/dataset/cc.py
@@ -27,9 +27,7 @@ from sec_certs.dataset.cc_eucc_common import (
     compute_heuristics_body,
     convert_all_pdfs_body,
     download_all_artifacts_body,
-    extract_all_frontpages,
-    extract_all_keywords,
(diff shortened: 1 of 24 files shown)


src/sec_certs/dataset/cc.py
```diff
@@ -395,24 +394,34 @@ def get_certs_from_web(
...
 
         logger.info(f"The resulting dataset has {len(self)} certificates.")
 
         if not keep_metadata:
             shutil.rmtree(self.web_dir)
 
+        if carry_processing_results:
+            self._carry_processing_results(old_certs)
+
         self._set_local_paths()
```
tkachyna: I would put `self._set_local_paths()` into an else branch as it is already called once inside the reconcile_document_states` fnc. The same applies for FIPS, EUCC, and PP datasets.

src/sec_certs/dataset/fips.py
```diff
@@ -294,8 +313,8 @@ def _extract_algorithms_from_policy_tables(self):
         )
```
tkachyna, line 315: just thinking, are we not missing here `self.update_with_certs(processed_certs)` as in other similar functions in this file?

jborsky: Yes, we do actually. It's mentioned and fixed in https://github.com/crocs-muni/sec-certs/pull/692.

PR #691, opened 2026-08-18
Add a document layer 

### LINKED ISSUE #542: Markdown headers issues with FIPS certificate ID extraction
Docling's Markdown output format uses # symbols to represent section headers (e.g., # Introduction ). This causes conflicts with FIPS certificate ID extraction rules that look for # to identify certificate identifiers.

The issue affects two pattern groups in the FIPS pipeline:

- fips_cert_id - Produces false positives
- fips_certlike - Causes unfeasible processing times due to excessive regex backtracking

So right now, the full FIPS would take hundreds of hours using Docling with the fips_certlike rule.

Proposed short-term fixes:
- Modify Docling's Markdown serialization to serialize headers without the hashtags
- Edit the affected regex rules to handle Markdown headers

I think the first one is better.

Long-term solution:
As part of my planned future enhancements, I plan to divide the rules to target different parts of the document (headers, footers, body content, etc.). For this, I will use Docling's JSON structure to search only in particular parts of the document. With this approach, the issue would be solved by not searching for cert IDs in section headers.

### LINKED ISSUE #570: 1.4. Tweak the integrated AI-enabled PDF parser using Docling

jborsky:
Second part of https://github.com/crocs-muni/sec-certs/issues/570. 

This PR adds a new document layer between the converter's output and the library features that needs some input from the artifacts. This closes https://github.com/crocs-muni/sec-certs/issues/542 and also prepares the layer for other features (such as hey, give me just text from footers/headers, hey, just give me tables, etc..). 

Summary of changes:
- Add a general `DocumentView` class that represents the view over the artifact.
- Implement the `get_full_text()` function for both Docling and pdftotext to return the complete unformatted text of the document. This is necessary for the current keyword extraction rules that require this specific format. For pdftotext, the functionality remains the same; it simply reads the text file. For Docling, the function returns plaintext output with unformatted tables, as table spacing has also been causing issues with the current regex rules.
- Wire the `get_full_text()` in the keyword extraction.

### DIFF
--- a/src/sec_certs/document/base.py
+++ b/src/sec_certs/document/base.py
@@ -32,4 +32,4 @@ class DocumentView(ABC):
         raise NotImplementedError("Not meant to be implemented by the base class.")
 
     def __str__(self) -> str:
-        return "/".join(self.artifact_path.parts[-4:])
+        return f"{type(self).__name__}({self.artifact_path})"
--- a/src/sec_certs/document/base.py
+++ b/src/sec_certs/document/base.py
@@ -2,13 +2,13 @@ from __future__ import annotations
 
 import logging
 from abc import ABC, abstractmethod
-from enum import StrEnum
+from enum import Enum
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
 
 
-class DocumentLayer(StrEnum):
+class DocumentLayer(str, Enum):
     BODY = "body"
     FURNITURE = "furniture"
 
--- a/src/sec_certs/document/base.py
+++ b/src/sec_certs/document/base.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+import logging
+from abc import ABC, abstractmethod
+from enum import StrEnum
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+class DocumentLayer(StrEnum):
+    BODY = "body"
+    FURNITURE = "furniture"
+
+
+class DocumentView(ABC):
+    @property
+    @abstractmethod
+    def artifact_path(self) -> Path:
+        """
+        Path to the file this view reads.
+        """
+        raise NotImplementedError("Not meant to be implemented by the base class.")
+
+    @abstractmethod
+    def get_full_text(self, layers: set[DocumentLayer] | None = None) -> str:
+        """
+        Get the whole unformatted text of the document
+
+        :param layers: content layers to include, all of them when None.
+        """
+        raise NotImplementedError("Not meant to be implemented by the base class.")
+
+    def __str__(self) -> str:
+        return "/".join(self.artifact_path.parts[-4:])
--- a/src/sec_certs/document/plaintext.py
+++ b/src/sec_certs/document/plaintext.py
@@ -11,16 +11,20 @@ logger = logging.getLogger(__name__)
 
 
 class PlainTextView(DocumentView):
-    def __init__(self, txt_path: Path):
+    def __init__(self, txt_path: Path, line_separator: str = LINE_SEPARATOR):
         self.txt_path = txt_path
+        self.line_separator = line_separator
 
     @property
     def artifact_path(self) -> Path:
         return self.txt_path
 
-    def get_full_text(self, layers: set[DocumentLayer] | None = None, line_separator: str = LINE_SEPARATOR) -> str:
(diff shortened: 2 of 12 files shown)


src/sec_certs/document/plaintext.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @property
+    def artifact_path(self) -> Path:
+        return self.txt_path
+
+    def get_full_text(self, layers: set[DocumentLayer] | None = None, line_separator: str = LINE_SEPARATOR) -> str:
+        if layers is not None and layers != set(DocumentLayer):
+            logger.debug(f"{type(self).__name__} has no notion of layers, returning the whole text of {self}.")
+
+        whole_text, _, _ = load_text_file(self.txt_path, -1, line_separator)
```
tkachyna: I am thinking about putting an info log here, that `get_full_text()` returns here the whole text no matter of what `layers` parameter contains.

src/sec_certs/document/plaintext.py
```diff
@@ -0,0 +1,26 @@
...
+
+class PlainTextView(DocumentView):
+    def __init__(self, txt_path: Path):
+        self.txt_path = txt_path
+
+    @property
+    def artifact_path(self) -> Path:
+        return self.txt_path
+
+    def get_full_text(self, layers: set[DocumentLayer] | None = None, line_separator: str = LINE_SEPARATOR) -> str:
```
tkachyna: I would put `line_separator` into the constructor as this method does not match its parents signature. 
```
class PlainTextView(DocumentView):
    def __init__(self, txt_path: Path, line_separator: str = LINE_SEPARATOR):
        self.txt_path = txt_path
        self.line_separator = line_separator
```

src/sec_certs/document/base.py
```diff
@@ -0,0 +1,35 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class DocumentLayer(str, Enum):
+    BODY = "body"
+    FURNITURE = "furniture"
+
+
+class DocumentView(ABC):
```
tkachyna, line 16: Do you plan add some tests for implementations of this base class in the future ?

tkachyna (commented):
Otherwise LGTM. :)

src/sec_certs/document/base.py
tkachyna: Now I am reading it in the previous PR, that tests will be added in follow up PRs. :D

PR #705, opened 2026-09-14
fix(dashboard): flatten nested fields for histogram, box and scatter charts

### LINKED ISSUE #685: Dashboard display error
**Describe the bug**
After adding a chart for the dashboard for FIPS-140, the chart presentation error is displayed:

Image

Here's how the chart has been set up:

Image

**Expected behavior**
If the problem is caused by the selection of x- (or y-) axis incompatible fields, I would expect to receive the error about this at the stage of editing the chart, not when it has been successfully saved and added to the dashboard.

**Additional context**
The more detailed error description can be accessed through Sentry.

xvalec01:
Fixes #685.

Histogram, box and scatter charts projected nested fields as `{"web_data.standard": 1}`. MongoDB returns this as a nested document, so the DataFrame had no `web_data.standard` column and rendering failed. Every nested field on these chart types was affected.

The pipeline now projects flattened names (`web_data_standard`), and `FigureBuilder.create_figure` reads them. This is the convention the aggregated chart types already use. Saved charts start working without being re-saved.

Unit tests added for the projection and the figure creation.


jborsky:
Thanks. I'll review it in the following days.

jborsky (approved):

